PR #12, opened 2024-02-15
Blueprint parsing experiment 1: Parsing as a form submission

adamziel:

🚧  Work in progress  🚧  

Blueprint parsing is just form submission! The data comes in, it’s validated, and gets loaded into new data structures.

This PR uses the with Symfony form library for that. Its input is the data class definition, like this:

```php
class UnzipStepInput {

	/**
	 * @Assert\NotBlank
	 * @Assert\NotNull
	 */
	public $zipFile;

	/**
	 * @Assert\Type("string")
	 * @Assert\NotBlank
	 * @Assert\NotNull
	 */
	public $toPath;
}
```

Then it takes a PHP array, validates that against the schema, outputs useful error messages (as opposed to JSON schema validators), and returns instances of the data classes populated with valid data

We could also generate the documentation pages and JSON schema from these data classes for convenient VS Code development.


adamziel:
Superseded by https://github.com/WordPress/blueprints/pull/14

PR #14, opened 2024-02-16
Blueprint parsing experiment 2: Generating PHP parser and data classes from a JSON Schema

adamziel:

This PR explores JSON Schema as the source of truth for Blueprints.

It uses the `swaggest/php-code-builder` package to generate:

* Model classes (plain old PHP objects)
* Builder classes (with validation and ->setLandingPage() etc methods)

So far, it looks really good:

* We have JSON Schema
* PHP won't diverge from JSON
* The complexity is mostly contained in code generation
* Code completion and static code analysis works like a charm
* We can trust the data we pass around, as opposed to arrays
* The footprint is tiny

 ## Extensibility

A plugin could define a new step ar a resource by adding it to the schema:

```
BlueprintBuilder::schema()->getProperties()->steps->items->anyOf[0]->oneOf[] = WPCLIStepBuilder::schema()
```

Tha API is clunky, sure, but we can expose convenient wrappers.

Defining new top-level properties isn't straightforward, but could be achieved by adding them via the reflections API. Maybe that isn't needed, though. There could be a dedicated `"plugins"` key where plugins would be free to define any schema they want without interfering with the main Blueprint structure. It's nice, because it would clearly separate core features from plugin features.

 ## Remaining work

* Changing the autogenerated file header


adamziel:
Merged in https://github.com/WordPress/blueprints/pull/15

PR #15, opened 2024-02-17
Port Step Handlers from the TypeScript library

adamziel:
Ports step handlers from the TypeScript library.

Embraces the approach from [Blueprint parsing experiment 2: Generating PHP parser and data classes from a JSON Schema](https://github.com/WordPress/blueprints/pull/14) and brings much of the Blueprint Parsing, Compilation, and Running from the TypeScript library.

There's more to do, still:

* Surfacing progress events
* EnableMultisite step
* Setting up WordPress

Much of this PR will likely be adjusted after merging. That being said, this API and methodology seems to be promising.


adamziel:
Even though this isn't nice and tidy, I'll go ahead and merge so the work on `trunk` can continue based on the ideas explored here.

PR #17, opened 2024-02-20
Mapping Experiment 1: JSON -> models with the jsonmapper library

adamziel:
An experiment at mapping a JSON blueprint definition to Blueprint PHP models.

The `swaggest` library used in trunk stores metadata in the models instances which makes debugging more difficult.

In these explorations, there is no data validation or error handling in this example. To validate, we'd still need to use a JSON schema validator like Opis.


adamziel:
The problem with this mapper library is it uses PHP comments and annotations to figure out the desired structure, and the swaggest code generator does not produce compatible annotations. It could be adjusted with enough work, but for now let's look for a solution that doesn't require that.

adamziel:

PR #18, opened 2024-02-20
Refactor and test rm and rmdir steps

reimic:
I've fused together the rm and rmdir steps. During testing I've noticed that the rmdir step uses `remove()` internally, which removes everything form the specified path - be it directories or files. While the rm's internal call to `unlink()` only handles files. Therefore, the rm step was essentially a method providing a subset of functionalities the rmdir provided. This seemed somewhat redundant.

Later on I realized the path resolving does not work on my windows machine. This was amended, also test verify both path types work fine.

Also, as discussed with @adamziel - the rm step is now more offensive in that is throws an exception if the provided path does not reference an existing file.

What does this PR do:
- removes the rmdir step entirely,
- adjusts the blueprints scheme,
- refactors the rm step to use `remove()` internally, allowing for deletion of directories and files alike,
- implements a todo in the rm step, allowing for a relative path as input,
- refactors the rm step to throw exceptions if the path does not reference an existing file,
- refactors the NativePHPRuntime to use a method from the Symphony Path API,
- adds tests for the refactored rm step,


tests/Unit/RmDirStepTest.php
```diff
@@ -0,0 +1,49 @@
...
+
+use WordPress\Blueprints\Model\Builder\RmDirStepBuilder;
+use WordPress\Blueprints\Runner\Step\RmDirStepRunner;
+
+it('should remove a directory', function () {
+    $path = __DIR__ . "/dir";
+    mkdir($path);
+    expect( file_exists($path)) -> toBeTrue();
+
+    $stepBuilder = new RmDirStepBuilder();
```
adamziel: Let's ditch testing the builders for now while they still settle and go directly with the model objects, e.g. `RmDirStep` in this case. So this example would be:

```suggestion
    $step = new RmDirStep();
    $step->path = $path;
```

tests/Unit/RmDirStepTest.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+use WordPress\Blueprints\Model\Builder\RmDirStepBuilder;
+use WordPress\Blueprints\Runner\Step\RmDirStepRunner;
+
+it('should remove a directory', function () {
+    $path = __DIR__ . "/dir";
```
adamziel: Let's give it a name that communicates the intention, say 
```suggestion
    $dirToRemove = __DIR__ . "/dir";
```

tests/Unit/RmDirStepTest.php
```diff
@@ -0,0 +1,49 @@
...
+    $step = new RmDirStepRunner();
+    $step->run($input);
+
+    expect( file_exists($path)) -> toBeFalse();
+});
+
+it ('should remove a directory with a subdirectory', function () {
+    $path = __DIR__ . "/dir";
+    $subpath = __DIR__ . "/dir/subdir";
+    mkdir($subpath, 0777, true);
```
adamziel: TIL mkdir can be recursive

tests/Unit/RmDirStepTest.php
reimic: Done.

composer.json
```diff
@@ -11,7 +11,8 @@
     "swaggest/php-code-builder": "^0.2.41"
   },
   "require-dev": {
-    "pestphp/pest": "^2.33"
+    "pestphp/pest": "^2.33",
+    "mockery/mockery": "^1.6.7"
```
adamziel: this can be removed

tests/Unit/RmDirStepTest.php
reimic: Done. But with the resolving... I am not certain it looks good now. WBY?

composer.json
reimic: Done!

adamziel: Thanks! You'll also need to run `composer remove mockery/mockery` to also remove them from the `composer.lock` file

src/WordPress/Blueprints/Runner/Step/RmStepRunner.php
```diff
@@ -2,18 +2,28 @@
...
+class RmStepRunner extends BaseStepRunner
+{
+    /**
+     * @param RmStep $input
+     */
+    function run(RmStep $input) {
+        $resolvedPath = $this->getRuntime()->resolvePath($input->path);
+        $fileSystem = new Filesystem();
+        if (false === $fileSystem->exists($resolvedPath)) {
+            throw new BlueprintException("Failed to remove the directory or file. No directory or file found at $resolvedPath", 0);
```
adamziel: Nitpick:
```suggestion
            throw new BlueprintException("Failed to remove \"$resolvedPath\": the directory or file does not exist.");
```

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
```diff
@@ -35,13 +35,57 @@ public function resolvePath( string $path ): string {
...
 		}
-		if ( $path[0] === '/' ) {
+		if ( $this->isAbsolutePath($path) ) {
 			return $path;
 		}
 
 		return join_paths( $this->getDocumentRoot(), $path );
 	}
 
+    private function isAbsolutePath(string $path): bool {
```
adamziel: Since it's ported from WordPress, let's add an attribution in the method docstring.

composer.json
reimic: Ah. I was unaware this won't happen on it's own. Thanks for the remark. Done now.

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
adamziel: This could use its own set of tests. Let's make it public for now and add a few tests to confirm it indeed works as expected. Later on we'll remove it to some Filesystem-related class.

Oh! Even better – it could be a function and not a method, in which case you'd add it to the `functions.php` file.

tests/Unit/RmStepTest.php
```diff
@@ -0,0 +1,118 @@
...
+    $this->step->run($input);
+
+    expect(file_exists($resolvedDirToRemove))->toBeFalse();
+});
+
+it ('should remove a directory with a subdirectory', function () {
+    $unresolvedSubDir = "dir/subdir";
+    $resolvedSubDir = $this->runtime->resolvePath($unresolvedSubDir);
+    mkdir($resolvedSubDir, 0777, true);
+    expect(file_exists($resolvedSubDir))->toBeTrue();
```
adamziel: This test could use `new FileSystem()` instance from Symfony and then it wouldn't need to check if `mkdir` failed since FileSystem throws an exception for us.

tests/Unit/RmStepTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+    $unresolvedDirToRemove = "dir";
```
adamziel: You could use the `dirname()` PHP function and avoid introducing new variables.

src/WordPress/Blueprints/Runner/Step/RmStepRunner.php
```diff
@@ -2,18 +2,28 @@
...
+        }
+        try {
+            $fileSystem->remove($resolvedPath);
+        } catch (IOException $exception) {
+            throw new BlueprintException("Failed to remove the directory or file at $resolvedPath", 0);
```
adamziel: Let's preserve the original error when re-throwing:

```suggestion
            throw new BlueprintException("Failed to remove the directory or file at $resolvedPath", 0, $exception);
```

src/WordPress/Blueprints/Runner/Step/RmStepRunner.php
reimic: Done. Also, adjusted tests.

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
reimic: On this note - there is a method in the FileSystem API that does the same thing. However, I have to review if its inner workings are actually identical.

See: Filesystem::isAbsolutePath

tests/Unit/RmStepTest.php
reimic: I added the assertion before the tested code execution for robustness. Although not entirely impossible, I doubt it could actually fail. So it mainly was a mean to highlight the change. Like: "see, it's there. now, we do sth. and it's gone. tada."

But I agree, this API is well thought out, and using it here can be beneficial.

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
reimic: Well, the FileSystem method seems less robust. But maybe it is sufficient? LMK which do you prefer @adamziel and I'll adjust accordingly.

On another note, Filesystem\Path::isAbsolute also has a method to handle this gracefully. It seems to be quite robust. Although, it might just be formatting... Still, I would lean towards using this method.

See: Filesystem\Path::isAbsolute

Also, please review if the method Filesystem\Path::makeAbsolute could be a viable alternative to most of the logic in the resolvePath method. I'll push a possible change, for you to review.

tests/Unit/RmStepTest.php
reimic: I can try if you insist. But `$unresolvedDirToRemove` is used twice. Once to resolve it for assertion sake, and also as input parameter path.

So, it is either, not using the variable, but invoking `dirname()` twice, or using the variable without additional methods. I would lean towards the latter. It looks slightly cleaner.

src/WordPress/Blueprints/Runner/Step/RmStepRunner.php
reimic: Done.

composer.lock
```diff
@@ -575,16 +575,16 @@
         },
         {
             "name": "swaggest/php-code-builder",
-            "version": "v0.2.41",
+            "version": "v0.2.42",
```
adamziel: There's still a bunch of unrelated changes in the composer.lock file. Let's keep it as it was.

src/WordPress/Blueprints/Runner/Step/RmStepRunner.php
```diff
@@ -2,18 +2,27 @@
...
+        if (false === $fileSystem->exists($resolvedPath)) {
+            throw new BlueprintException("Failed to remove $resolvedPath: the directory or file does not exist.");
+        }
+        try {
+            $fileSystem->remove($resolvedPath);
+        } catch (IOException $exception) {
+            throw new BlueprintException("Failed to remove the directory or file at $resolvedPath", 0, $exception);
+        }
+    }
+}
```
adamziel: Nitpick: there's no empty line at the end of the file which GitHub marks like this:

CleanShot 2024-02-21 at 14 42 35@2x

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
```diff
@@ -32,14 +33,7 @@ public function getDocumentRoot(): string {
...
 	public function resolvePath( string $path ): string {
-		if ( ! strlen( $path ) ) {
-			return $this->getDocumentRoot();
-		}
-		if ( $path[0] === '/' ) {
-			return $path;
-		}
-
-		return join_paths( $this->getDocumentRoot(), $path );
+        return Path::makeAbsolute($path, $this->getDocumentRoot());
```
adamziel, line 36: Nice! Just a nitpick – let's indent this line. Code formatting is still messy in general and this repo needs a formatter, but let's keep at least the indentation in tact.

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
```diff
@@ -135,5 +129,4 @@ public function startProcess(
 			$timeout
 		);
 	}
-
-}
+}
```
adamziel: Nitpick: no empty newline

tests/Unit/RmStepTest.php
```diff
@@ -0,0 +1,117 @@
...
+    $this->fileSystem = new Filesystem();
+});
+
+it('should remove a directory when an absolute path is provided', function() {
+    $unresolvedDirToRemove = "dir";
+    $resolvedDirToRemove = $this->runtime->resolvePath($unresolvedDirToRemove);
+    $this->fileSystem->mkdir($resolvedDirToRemove);
+
+    $input = new RmStep();
+    $input->path = $resolvedDirToRemove;
```
adamziel: The step should accept a relative path, not a resolved absolute path. Does this test pass? If so, is this a bug?

adamziel: Aha, this specifically tests what happens when an absolute path is provided. Let's make that apparent in the code structure, e.g. by collapsing these two lines into one:

```diff
-    $unresolvedDirToRemove = "dir";
-    $resolvedDirToRemove = $this->runtime->resolvePath($unresolvedDirToRemove);
+    $resolvedDirToRemove = $this->runtime->resolvePath("dir");
```

composer.lock
reimic: I'll rollback the whole file. Yet I fail to see how I have cause this? I did run the code suggested in the README file. Why would it not adhere to the specs in the `composer.json` file?

tests/Unit/RmStepTest.php
adamziel: I guess it's a matter of taste at this point. Personally I'd just call `dirname()` twice to avoid introducing additional names for things.

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
reimic: Yeah. I've noticed that after the push. But - YES, PLEASE, a formatter is a must.

tests/Unit/RmStepTest.php
```diff
@@ -0,0 +1,117 @@
...
+    $input->path = $unresolvedDirToRemove;
+
+    $this->step->run($input);
+
+    expect($this->fileSystem->exists($resolvedDirToRemove))->toBeFalse();
+});
+
+it ('should remove a directory with a file', function () {
+    $unresolvedFile = "dir/file.txt";
+    $resolvedFile = $this->runtime->resolvePath($unresolvedFile);
```
adamziel: I know I asked the opposite earlier, but perhaps these names would be easier to understand after all:

```suggestion
    $relativePath = "dir/file.txt";
    $absolutePath = $this->runtime->resolvePath($relativePath);
```

I found myself thinking "what's 'resolved'?" when reading that code fragment and I'd rather wonder "what do we need a relative path for?"

tests/Unit/RmStepTest.php
reimic: Haven't we agreed on it accepting any? Internally resolving unresolved paths, and letting absolute ones be? Could you describe the ideal behavior you expect from this step?

tests/Unit/RmStepTest.php
```diff
@@ -0,0 +1,117 @@
...
+
+   $input = new RmStep();
+   $input->path = $unresolvedFileToRemove;
+
+   $this->step->run($input);
+
+   expect($this->fileSystem->exists($resolvedFileToRemove))->toBeFalse();
+});
+
+it ('should throw an exception when asked with a relative path to remove a nonexistent directory ', function() {
```
adamziel:
```suggestion
it ('should throw an exception when asked to remove a non-existent directory (using a relative path) ', function() {
```

tests/Unit/RmStepTest.php
```diff
@@ -0,0 +1,117 @@
...
+    $dirToRemove = "dir";
+
+    $input = new RmStep();
+    $input->path = $dirToRemove;
+
+    $resolvedDirToRemove = $this->runtime->resolvePath($dirToRemove);
+    expect(fn() => $this->step->run($input))->toThrow(BlueprintException::class, "Failed to remove $resolvedDirToRemove: the directory or file does not exist.");
+});
+
+it ('should throw an exception when asked with an absolute path to remove a nonexistent directory', function () {
```
adamziel:
```suggestion
it ('should throw an exception when asked to remove a non-existent directory (using an absolute path)', function () {
```

tests/Unit/RmStepTest.php
reimic: As a general rule I would suggest to move from `resolved` -> `absolute` and `unresolved` -> `relative`. This would also apply to the `resolvePath` method. Maybe `toAbsolutePath` (as in Path util) would work best?

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
adamziel: > Well, the FileSystem method seems less robust. But maybe it is sufficient? 

I don't have enough context here, where does it fall short compared to the one from WordPress?

adamziel: > Also, please review if the method Filesystem\Path::makeAbsolute could be a viable alternative to most of the logic in the resolvePath method. I'll push a possible change, for you to review.

Nice, you're right!

tests/Unit/RmStepTest.php
adamziel: > Although not entirely impossible, I doubt it could actually fail

It's more common than you'd think – read-only filesystems, already existing directories, invalid paths and other corner cases happen all the time.

composer.lock
adamziel: Here's what I think happened:

1. You installed a new composer package, which also updated some of the dependencies
2. You uninstalled that package, which did not roll back the other changes in the lock file

The changes seem fine, actually, I'd just leave them out from this PR as they're completely unrelated.

tests/Unit/RmStepTest.php
adamziel: Yes we did, I realized after posting the initial comment. My second comment in this thread contains a suggestion to make the test slightly less confusing.

tests/Unit/RmStepTest.php
adamziel: We're testing using a `NativePHP` runtime so absolute and resolved are synonymous, but the runtime could eventually be something like a remote Docker container and then the resolved path wouldn't exist on a local computer. `toAbsolutePath` doesn't quite express the intention. But perhaps remote runtimes won't make it to the final library release? Let's leave that method name as it is for now, but adjust the variable names in the test cases.

tests/Unit/RmStepTest.php
reimic: Let me know if you agree, but I'd stick with `nonexistent`. This is a nitpick, really. Historically hyphens were used in BrE, and a single word in AmE. Generally, the trend is to accept nonhyphenated words in BrE, too. Could we stick to the AmE spelling, as we use behavior instead of behaviour.

The other thing - sure. Will do.

tests/Unit/RmDirStepTest.php
reimic: Renaming to agreed upon convention.

tests/Unit/RmStepTest.php
reimic: I acquiesce.

src/WordPress/Blueprints/Runner/Step/RmStepRunner.php
reimic: Done.

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
reimic: Done.

tests/Unit/RmStepTest.php
reimic: Done. 

Not relevant to this pr, but as food for thought  -> this is an internal method, if a certain name would work best at this point in time, and there are no clear plans that definitely invalidate it. Maybe we should do what is best now, and adjust accordioning when a true need arises?

tests/Unit/RmStepTest.php
reimic: Done.

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
reimic: I've noticed this displays properly locally and only looks weird here.

adamziel: It's probably due to mixing tabs and spaces

tests/Unit/RmStepTest.php
```diff
@@ -0,0 +1,116 @@
...
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Path;
+use WordPress\Blueprints\BlueprintException;
+use WordPress\Blueprints\Model\DataClass\RmStep;
+use WordPress\Blueprints\Runner\Step\RmStepRunner;
+use WordPress\Blueprints\Runtime\NativePHPRuntime;
+
+
+beforeEach(function() {
+    $this->documentRoot = Path::makeAbsolute("test", sys_get_temp_dir());
```
adamziel, line 12: Nice

tests/Unit/RmStepTest.php
adamziel: Makes sense to me 👍

PR #19, opened 2024-02-20
Mapping Experiment 2: JSON -> models with Jane-php library and Symfony normalizers

adamziel:
This explores generating PHP classes using the Jane PHP library:

https://jane.readthedocs.io/en/latest/documentation/JsonSchema.html

On the upside, it creates models, normalizers, and validators which is all we need to go from JSON and Builder objects to CompiledBlueprint. The validators are statically declared and don't rely on parsing and validating data usint the JSON schema.

On the downside:
* The generated model classes have an `$initialized` property which we don't need but the normalizers seem to rely on it.
* Validating collections like blueprint.steps doesn't see to be covered by the generated validators


generated/Normalizer/BlueprintNormalizer.php
```diff
@@ -0,0 +1,560 @@
...
+            }
+            if (\array_key_exists('steps', $data)) {
+                $values_3 = [];
+                foreach ($data['steps'] as $value_4) {
+                    $value_5 = $value_4;
+                    if (is_array($value_4) and isset($value_4['step']) and isset($value_4['slug'])) {
+                        $value_5 = $this->denormalizer->denormalize($value_4, 'WordPress\\Blueprints\\Generated\\Model\\ActivatePluginStep', 'json', $context);
+                    } elseif (is_array($value_4) and isset($value_4['step']) and isset($value_4['slug'])) {
+                        $value_5 = $this->denormalizer->denormalize($value_4, 'WordPress\\Blueprints\\Generated\\Model\\ActivateThemeStep', 'json', $context);
+                    } elseif (is_array($value_4) and isset($value_4['step']) and isset($value_4['fromPath']) and isset($value_4['toPath'])) {
```
adamziel: This pattern matching requires all the `Step` keys to be in place, not just the `step` property. It will not produce useful error messages on incomplete input.

Also, it uses `and` 😱

generated/Validator/BlueprintConstraint.php
```diff
@@ -0,0 +1,11 @@
...
...null.'])]), 'plugins' => new \Symfony\Component\Validator\Constraints\Optional([new \Symfony\Component\Validator\Constraints\Type(['0' => 'array']), new \Symfony\Component\Validator\Constraints\NotNull(['message' => 'This value should not be null.'])]), 'siteOptions' => new \Symfony\Component\Validator\Constraints\Optional([new \Symfony\Component\Validator\Constraints\NotNull(['message' => 'This value should not be null.'])]), 'phpExtensionBundles' => new \Symfony\Component\Validator\Constraints\Optional([new \Symfony\Component\Validator\Constraints\Type(['0' => 'array']), new \Symfony\Component\Validator\Constraints\NotNull(['message' => 'This value should not be null.'])]), 'steps' => new \Symfony\Component\Validator\Constraints\Optional([new \Symfony\Component\Validator\Constraints\Type(['0' => 'array']), new \Symfony\Component\Validator\Constraints\NotNull(['message' => 'This value should not be null.'])]), '$schema' => new \Symfony\Component\Validator\Constraints\Optional([new \Symfony\Component\Validator\Constraints\Type(['0' => 'string']), new \Symfony\Component\Validator\Constraints\NotNull(['message' => 'This value should not be null.'])])], 'allowExtraFields' => false])];
```
adamziel: It doesn't seem to check whether the `step` entries are indeed instances of the Step class

```
'steps' => Constraints\Optional([
    new Constraints\Type(['0' => 'array'])
])
```

adamziel:
Downside of this approach: it requires including bulky Symfony components in the final `.phar` downloaded by Playground.

adamziel:
I confirmed the type of nested objects is indeed not validated. This faulty Blueprint validates perfectly well:

```json
{
	"plugins": [
		["akismet"]
	]
}
```

Only when I manually added the following constraints, the Validator detected that `plugins` entries aren't supposed to be arrays:

```php
new \Symfony\Component\Validator\Constraints\All( [
	new \Symfony\Component\Validator\Constraints\AtLeastOneOf( [
		new \Symfony\Component\Validator\Constraints\Type( [ '0' => 'string' ] ),
		new \Symfony\Component\Validator\Constraints\Type( [ '0' => UrlResource::class ] ),
	] ),
] ),
```

Symfony validators produced the following error message:

```
Blueprint[plugins][0]: This value should satisfy at least one of the following constraints: [1] This value should be of type string. [2] This value should be of type WordPress\Blueprints\Generated\Model\UrlResource."
```

Which is great, but also not that helpful if the developer provided JSON as input. This tells me we may need two stages of validation:

1. JSON validation to check if it conforms to the schema
2. Blueprint object validation to confirm it was constructed properly

Here's a situation where that would matter:

```php
$blueprint = Blueprints::fromJSON($jsonData);
$blueprint->addPlugin( ["invalid data – this is not supposed to be an array"] );
Blueprints::run($blueprint);
```

adamziel:
`jane` library produces models with private properties that expose getters and setters. The data normalization/mapping works as follows:

```php
$object = new Blueprint();
$object->setWpVersion( $data['wpVersion'] );
```

It may seem fine, but if the `wpVersion` property is required, the Blueprint `$object` is invalid by default and only made valid by further processing. This will make breaking changes difficult to detect and migrate:

```php
(new Blueprint())->setWpVersion('6.4')->setSteps([
     (new RmStep())->setPath('/wordpress/test.txt'),
     (new InstallPluginStep()) // Ooops, "slug" is required but not set and there's nothing to detect that
]);
```

I'd love a model that cannot express an invalid state, e.g.:

```php
Blueprint::create([
    'wpVersion' => '6.4',
    'steps' => [
        RmStep::create([ 'path' => '/wordpress/test.txt' ]),
        InstallPluginStep::create([]) // Throws an exception – the required "slug" option wasn't provided
    ]
]);
```

It may be less pretty than the builder pattern, but with proper PHPDoc annotations it could still offer autocompletion AND it would save developers a ton of debugging.

That last example may not seem too different from just writing a JSON-like array directly:

```php
$blueprint = [
    'wpVersion' => '6.4',
    'steps' => [
        [ 'step' => 'rm', 'path' => '/wordpress/test.txt' ],
        [ 'step' => 'installPlugin' ]
    ]
];
```

The difference is that in the former case we'd get an exception the moment we try to construct an invalid model instance. On the other hand, in the latter case perhaps we could get PHPStan or Psalm to provide the same kind of instant feedback?


Alternatively, the builder-created Blueprint could go through a second phase of JSON Schema validation before it's processed in any way.

adamziel:
Another two issues with this library:

* It doesn't provide default values for discriminator properties. For example, `CpStep::$step` isn't set to "cp" but it must be populated by the caller.
* It generates two versions of every Normalizer class and uses a conditional to load one of them – why would it do that?

There are no extension points so using `Jane` as it is looks like a no-go. It might, however, be possible to build a custom code generator using the library.

adamziel:
I used Jane as a library to generate the models just the way we need them in https://github.com/WordPress/blueprints/pull/19/commits/ddb30d7c58e1ce34fe0fff9173c1e32f2431e4bb. Mapping from parsed JSON objects is implemented via https://jsonmapper.net/. The last missing bit is adding JSON Schema validation via https://opis.io/ and removing the swaggest library.

Example generated model:

```php
class InstallPluginStep implements StepDefinitionInterface
{
	public const DISCRIMINATOR = 'installPlugin';

	/** @var Progress */
	public $progress;

	/** @var bool */
	public $continueOnError;

	/**
	 * The step identifier.
	 * @var string
	 */
	public $step = 'installPlugin';

	/** @var string|ResourceDefinitionInterface */
	public $pluginZipFile;

	/** @var InstallPluginOptions */
	public $options;
}
```

adamziel:

PR #20, opened 2024-02-20
Mapping Experiment 3: JSON -> models with the php-json-schema-model-generator library

adamziel:
Explores this library for generating and mapping models:

https://php-json-schema-model-generator.readthedocs.io/en/latest/gettingStarted.html#installation

The output isn't compelling:

* The names are mangled
* Models have a bunch of logic in them
* Setters attempt to validate the incoming data
* Validation is coupled and co-located with the data object

I don't think we'll go with this library


adamziel:
I'm closing this PR – the library isn't compelling compared to the alternatives.

PR #21, opened 2024-02-20
Mapping Experiment 4: Custom JSON Schema -> PHP models code generator

adamziel:
This is an attempt to replace opinionated code generation and mapping libraries with an equally opinionated, custom, and small code generator.

It started simple, but seems like a large project on its own as it requires handling anyOf, allOf, nested objects, consts, discriminators, mixed types, additionalProperties, and other JSON Schema features.


adamziel:

PR #22, opened 2024-02-20
Mapping Experiment 5: Generate "Valid" and "Dirty" model classes

adamziel:
An attempt to store the internal parsed Blueprint state without the internal swaggest state, see https://github.com/WordPress/blueprints/issues/13 for more details.

There is indeed a degree of separation, but now all the runner classes must accept a `Valid` model instance like `ValidCpStep` instance, which is weird.

Perhaps I shouldn't try so hard to make "valid" a static property of the data type like I did in TypeScript, but define it as an internal contract. There is just one Blueprint class, and it's valid if it passed the `validate()` check. There's no data rewriting or class dichotomy.


adamziel:

PR #23, opened 2024-02-22
Mapping Experiment 6: Blueprints parser, mapper, and validator autogenerated from JSON Schema

adamziel:

Solves the Blueprint parsing problem as follows:

* JSON Schema is the source of truth for the Blueprint data structure
* Model classes are generated from JSON Schema
* Input Validation is done against the JSON Schema
* Mapping validated input into model instances is indirectly done based on the JSON Schema

## Summary of the approach

We use [Janephp](https://github.com/janephp/janephp/) to infer the PHP model class structure from the JSON schema

We then use Jane's output to generate PHP code using [Nette PHP generator](https://github.com/nette/php-generator):

* Model classes
* Interfaces for groups of related classes: `StepDefinitionInterface`, `ResourceDefinitionInterface`.
* Interface resolution map for the [JsonMapper library](https://jsonmapper.net/).
* PHP docstrings with more accurate types than Jane to guide the mapping process

From there, the data pipeline looks as follows:

1. Parse raw JSON
2. Validate it with Opis
3. Map it into PHP models using JsonMapper

Or, if you're consuming the PHP API directly:

1. Create Model instances
2. Validate them with Opis

## Remaining work

- [x] Solve merge conflicts

## Rationale

Parsing and mapping JSON [in TypeScript is a simple problem, but in PHP it's surprisingly involved](https://github.com/WordPress/blueprints/issues/13#issuecomment-1956607670). This PR combines three previously explored approaches into a single pipeline:

* https://github.com/WordPress/blueprints/pull/17
* https://github.com/WordPress/blueprints/pull/19
* https://github.com/WordPress/blueprints/pull/21


adamziel:
Everything in the Blueprints library hinges on this part and now it's there, yay! Let's merge!

adamziel:
Here's how this PR fits [the original design goals](https://github.com/WordPress/blueprints/issues/13):

> Extensibility – At minimum, plugins should be able to register new steps and resource handlers. At maximum, they could register new top-level properties and extend the existing semantics.

While there are no utilities to register new steps yet, the architecture supports doing that in three steps:

1. Extend the JSON schema – later on we'll build it to a PHP array or class and expose WordPress-like filters.
2. Adjust the Step list passed to the mapper – this can be easily wired in the Container builder once we have filters.
3. Registering a new runner in the container – ditto.

I no longer think registering top-level properties is a good idea. Instead, let's just expose a top-level key like `"extensions"` and provide a dedicated namespace for each extension.

> Clarity – Even with a few plugins, it should be easy to tell what is the current Blueprint definition and where each part comes from.

With just core in place, JSON Schema is the source of truth. Any additional steps and resources will come from plugins. They may have to be namespaced just to make their origin clearer, e.g. `woo/import-products` instead of just `import-products`.

> Static analysis – Blueprints should have a JSON schema and, as a PHP library, IDE support for autocompletion.

✅ JSON Schema
✅ Statically declared PHP classes with autocompletion support

> Small footprint – The parser should be as small as it can reasonably be, without involving dozens of MBs of libraries.

The only two libraries involved in parsing are:

* Opis JSON Schema validator – there's no getting around that, we need a validator library regardless of the parsing technique.
* JSON Mapper – same as above. The library is relatively small BUT it adds a bunch of dependencies. Later on, we may need to replace it with a smaller custom implementation. For now, let's keep it around.

adamziel:
Furthermore, we have a convenient builder class for PHP developers:

```php
$blueprint = BlueprintBuilder::create()
	->withWordPressVersion( 'https://wordpress.org/latest.zip' )
	->withSiteOptions( [
		'blogname' => 'My Playground Blog',
	] )
	->withWpConfigConstants( [
		'WP_DEBUG'         => true,
		'WP_DEBUG_LOG'     => true,
		'WP_DEBUG_DISPLAY' => true,
		'WP_CACHE'         => true,
	] )
	->withPlugins( [
		'https://downloads.wordpress.org/plugin/hello-dolly.zip',
		'https://downloads.wordpress.org/plugin/gutenberg.17.7.0.zip',
	] )
	->withTheme( 'https://downloads.wordpress.org/theme/pendant.zip' )
	->withContent( 'https://raw.githubusercontent.com/WordPress/theme-test-data/master/themeunittestdata.wordpress.xml' )
	->withSiteUrl( 'http://localhost:8081' )
	->andRunSQL( <<<'SQL'
		CREATE TABLE `tmp_table` ( id INT );
		INSERT INTO `tmp_table` VALUES (1);
		INSERT INTO `tmp_table` VALUES (2);
		SQL
	)
	->withFile( 'wordpress.txt', 'Data' )
	->toBlueprint();
```

PR #25, opened 2024-02-22
Downgrade to PHP 7.4, move from Pest to PHPUnit, establish CI

reimic:
### What is this PR doing?
Acts as an intermediary step in reaching goals set in #26 , #27.

- Downgrades the lib to PHP 7.4
- Adds 'Check PHP CS == 7.4' step to GitHub Actions
- Replaces Pest with PHPUnit
- Rewrites RmStep tests in PHPUnit 
- Adds matrix unit tests step for all operating systems and all PHP versions >= 7.4

### Caveats
- Dependencies versions are wildcards. This is inherently unstable, but our true goal is to have stable 7.0 support. Establishing fixed versions is a tedious job that does not seem crucial at this temporary step.
- Unit tests should include `prefer-lowest` as one of the arguments in the `dependency_version` list. This will allow for the broadest version support. But at this point in time it fails at 7.4. Fixing this serves little purpose as, see logic above.
- Although we know PHPUnit 6.5 should be used, since this is the newest versions supporting PHP == 7.0, PHPUnit 9.6 is used here. This is due to all lower versions throwing the pesky `PHP Fatal error:  Cannot acquire reference to $GLOBALS`. This is still a step in the right direction as writing PHPUnit tests in 9.6 and then downgrading them to 6.5 is still easier, than writing in Pest and then going for PHPUnit 6.5.
- It should be analyzed if it is in fact beneficial to run the matrix with each pull/pr. Alternatives: cron, only pr and more.


reimic:
Coding standards matching WordPress' support for 7.0:
- downgrade the whole lib to 7.0,
- implement a GitHub actions step checking for 7.0 coding standards:
  - this is surprisingly not straight forward, since many quality tools require a greater PHP version,
  - older version exist, but why use a tool in a version that is no longer supported, when supported tools that handle PHP 7.0 exist?
  - PHP CS Fixer requires >=7.4, [source](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer?tab=readme-ov-file#supported-php-versions)
  - Laravel Pint is build on PHP CS Fixer, [source](https://laravel.com/docs/10.x/pint#introduction)
  - PHPStan requires >=7.2, [source](https://phpstan.org/user-guide/getting-started#:~:text=Sponsor%20PHPStan!-,Getting%20Started,-PHPStan%20requires%20PHP)
  - CodeSniffer seems ok, also is integrated into PHPStorm [source](https://github.com/PHPCSStandards/PHP_CodeSniffer/#requirements)
  - Mess Detector also seems ok, also is integrated into PHPStorm [source](https://phpmd.org/download/index.html#:~:text=composer.phar%20install-,Requirements,-PHPMD%20itself%20is)
  
 I'll explore CodeSniffer and Mess Detector more. Any takes @adamziel ?

adamziel:
> downgrade the whole lib to 7.0,

There's a lot happening here. Let's ship this one with failing tests and downgrade in a separate PR.

> this is surprisingly not straight forward, since many quality tools require a greater PHP version,

That's fine. People should be able to run the published Blueprints package in their PHP 7 codebase, but contributing to the project doesn't have the same constraint. It's perfectly fine for the developer tools to require PHP 8. Test runner is more tricky since we want the tests to pass on PHP 7 so, presumably, the runner should be compatible with that version. If that means we need to move from Pest to PHPUnit, then let's do that.

> CodeSniffer seems ok, also is integrated into PHPStorm [source](https://github.com/PHPCSStandards/PHP_CodeSniffer/#requirements)

This is the standard WordPress tool, let's go with that one.

adamziel:
I wonder if composer can even enforce PHP 7.0 in the project while allowing the dev tools to run on PHP 8.2 🤔 Perhaps we'll need a separate list of global composer dependencies just to free them from these restrictions.

reimic:
> If that means we need to move from Pest to PHPUnit, then let's do that.

It does. I'll start with my recent pr. For PHP 7.0 the valid PHPUnit version is 6.5. [source](https://phpunit.de/manual/6.5/en/installation.html#installation.requirements) Of course it is no longer supported. But this is expected.

What is more, this would be beneficial for cross-version and cross-os testing, with something along these lines:
```yml
tests:  
    runs-on: ${{ matrix.os }}  
    strategy:  
      fail-fast: true  
      matrix:  
        os: [ubuntu-latest, macos-latest, windows-latest]  
        php: ['7.0', '7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3']  
        dependency_version: [prefer-lowest, prefer-stable]  
```

> I wonder if composer can even enforce PHP 7.0 in the project while allowing the dev tools to run on PHP 8.2

Not with a trivial solution. The question being - which tolls lack a acceptable 7.0 alternative?

reimic:
It's quite confusing. For PHP 7.0 `"symfony/filesystem":"^3.4"` is the highest valid version. But it conflicts with `"symfony/http-kernel"` which in turn has to be downgraded to `"symfony/http-kernel": "^3.4.49"` which is incompatibile with a buch of stuff. Among others `"symfony/http-client"` which has no PHP 7.0 compabile version at all.

For now I'll do a workaround and focus on moving from `Pest` to `PHPUnit`. But those will use Filesystem in it's current version, which will be downloaded later. I guess this is the pinnacle of iterative development...

reimic:
Let's continue in #30 and focus on a basic pipeline and Pest -> PHPUnits change.

reimic:
Let's continue in #30 and focus on a basic pipeline and Pest -> PHPUnits change.

reimic:
Well. After lengthy discussions with @adamziel it became apparent that downgrading / testing are more interconnected than initially expected. See problems mentioned at #30.

So, to meet #27 and finally #26, we'll downgrade the code as far back as possible without extraneous effort. Essentially to PHP 7.4. At this version we can rewrite / keep writing unit tests in PHPUnit 6.5 and have pipelines run somewhat smoothly. 

And after that is done, and the code mostly works 👼, we can carefully move to 7.0. Which will include replacing or transpiling dependencies.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+
+on:
+  push:
+    branches:
+    - trunk
+  pull_request:
+
+jobs:
+  php-cs-check:
+    name: 'Check PHP CS == 7.4'
```
adamziel: The name suggests this checks whether the PHPCS version is `7.4` but I think the intention was to:
```suggestion
    name: 'PHP 7.4 compatibility (via PHPCS)'
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          tools: composer:v2
+          coverage: none
+
+      - name: Install PHP dependencies
+        run: composer update --${{ matrix.dependency_version }} --no-interaction --prefer-dist --optimize-autoloader
+
+      - name: Tests
```
adamziel:
```suggestion
      - name: Run tests
```

phpunit.xml
```diff
@@ -1,17 +1,8 @@
...
-         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.3/phpunit.xsd"
-         bootstrap="vendor/autoload.php"
-         colors="true"
->
+<phpunit bootstrap="vendor/autoload.php" colors="true">
     <testsuites>
-        <testsuite name="Test Suite">
-            <directory suffix="Test.php">./tests</directory>
+        <testsuite name="tests">
+            <directory>tests</directory>
```
adamziel, line 5: Why remove the Test.php suffix requirement?

src/WordPress/Blueprints/BlueprintParser.php
```diff
@@ -152,14 +153,15 @@ private function findSubErrorForSpecificAnyOfOption( \Swaggest\JsonSchema\Except
...
-			return;
+			return $e->error;
 		}
 		foreach ( $e->subErrors as $subError ) {
 			$subError = findSubErrorForSpecificAnyOfOption( $subError, $anyOfRef );
 			if ( $subError !== null ) {
 				return $subError;
 			}
 		}
+        return $e;
```
adamziel: Nitpick: This is indented with spaces while the rest of this file is indented with tabs

src/WordPress/Blueprints/BlueprintParser.php
```diff
@@ -152,14 +153,15 @@ private function findSubErrorForSpecificAnyOfOption( \Swaggest\JsonSchema\Except
 			}
 		}
 		if ( ! $e->subErrors ) {
-			return;
+			return $e->error;
```
adamziel: I'm not sure if that's right, but this function isn't used at the moment so you could comment it out as well. We'll soon need some version of it that can work with the Opis JSON Schema validation library (this version works with the no-longer-used Swaggest library).

src/WordPress/Blueprints/ResourceManager/Resolver/FilesystemResourceResolver.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace WordPress\Blueprints\ResourceManager\Resolver;
```
adamziel: Why change the namespace from `Resource` to `ResourceManager`? Is `Resource` not allowed in PHP 7.4? If so, let's use `Resources` instead as resolvers are not manager-specific.

adamziel: Also, this PR shows a lot of deleted files and a lot of created files, but I think they are the same files moved to a new location. Check out the `git mv` command – it should make this PR a lot easier to read.

src/WordPress/Blueprints/ResourceManager/Resolver/FilesystemResourceResolver.php
```diff
@@ -0,0 +1,41 @@
...
+class FilesystemResourceResolver implements ResourceResolverInterface
+{
+
+    public function parseUrl(string $url): ?ResourceDefinitionInterface
+    {
+        if (! str_starts_with($url, 'file://')) {
+            return null;
+        }
+
+//    TODO does this Builder exist?
```
adamziel: Nope, good catch. It's just `FilesystemResource` now

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
```diff
@@ -10,11 +10,13 @@
 class NativePHPRuntime implements RuntimeInterface {
 
 	public Filesystem $fs;
+    protected string $documentRoot;
```
adamziel: Tabs vs spaces again. Would you be able to also set up PHPCBF to auto-format all those files?

.github/workflows/ci.yml
reimic: `CS` was for `Coding Standards`, but your option leaves no room for misunderstandings. Also, this is a whoosh moment for me. I did not realize it could mean `Code Sniffer`, too. Fixed.

.github/workflows/ci.yml
reimic: Sure. Why not. Fixed.

phpunit.xml
reimic: Works either way. Less is more.

src/WordPress/Blueprints/BlueprintParser.php
reimic: The PHPStorm autolinter seems to be underconfigured. As this is not the only comment regarding linting, there is a choice for you here - option a. we close our eyes, accept tiny linting mishaps, and configure and run a linter through the whole project in a separate PR, b. we do it now, and make this a massive "set all standards at once PR". I can go either way. The choice is yours.

src/WordPress/Blueprints/BlueprintParser.php
reimic: I am commenting out the unused method and leaving several TODOs hinting at things to consider.

src/WordPress/Blueprints/ResourceManager/Resolver/FilesystemResourceResolver.php
reimic: Point 1: Resource -> ResourceManager -> Resources
Yeah. It's a soft reserved word. [source](https://www.php.net/manual/en/reserved.other-reserved-words.php) It isn't a must, but it's a tiny ask and we can abide it. Nonetheless, I am changing the name to `Resources` as per your request.

reimic: Point 2: The change was a tip of the IDE and I deemed it reasonable. But it was done as an suggested action in PHPStorm, so it did its own thing. 

I wonder if it can be configured to use git mv internally... But, that is beyond the point. At this stage - will moving it back and then here again, fix the way it is displayed in this PR?

src/WordPress/Blueprints/ResourceManager/Resolver/FilesystemResourceResolver.php
reimic: Cool. Review if the proposed fix is suitable, please.

src/WordPress/Blueprints/Runtime/NativePHPRuntime.php
reimic: [see this response](https://github.com/WordPress/blueprints/pull/25/files/98fdb7dc60d43742bb430ca937d96bf3e69682de#r1503291952)

src/WordPress/Blueprints/Runner/Step/CpStepRunner.php
```diff
@@ -2,18 +2,53 @@
...
+            // cp copies all files in the source directory to the destination directory,
+            // creating any files or directories needed. This mode of operation requires an additional option flag,
+            // typically r, to indicate the recursive copying of directories.
+            // If the destination directory already exists, the source is copied into the destination,
+            // while a new directory is created if the destination does not exist.
+
+            // TODO establish how $input->toPath should be handled
+            // should copy to document root when null or when ""?
+
+            $fileSystem->copy($resolvedFromPath, $resolvedToPath, true);
```
reimic: I've pondered it for a moment or two. And I think the right way forward here is to implement our own copy method. 

My reasoning: 
- As a user, I'd see a lot of value in using it the way unix specifies it. [source](https://en.wikipedia.org/wiki/Cp_(Unix)#:~:text=Meyering.%5B5%5D-,Operating%20modes,-%5Bedit%5D) 
- Would work more akin to RmStep, handling both files and directories alike.
- We intend to have a limited amount of steps, so it's not like we'll have a ton of custom methods to maintain.
- We'll be less reliant on Filesystem, which we must downgrade to v3.x for PHP 7.0.

src/WordPress/Blueprints/Runner/Step/BaseStepRunner.php
```diff
@@ -2,33 +2,39 @@
 
 namespace WordPress\Blueprints\Runner\Step;
 
-use WordPress\Blueprints\Context\ExecutionContext;
-use WordPress\Blueprints\Resource\ResourceManager;
+// TODO - what about the ExecutionContext?
```
adamziel: It's the former name of `NativeRuntime`, it's safe to ignore/remove for now

src/WordPress/Blueprints/Runner/Step/CpStepRunner.php
adamziel: What would a custom method enable that the `$fileSystem->copy()` method doesn't?

tests/Unit/stream_str_replace_Test.php
```diff
@@ -1,39 +1,39 @@
...
-	rewind( $source );
-
-	$generator = stream_str_replace( $source, "https://example.com", "https://playground.internal" );
-	$chunks    = iterator_to_array( $generator );
-	$expected  = str_replace( "https://example.com", "https://playground.internal", $data );
-
-	expect( implode( "", $chunks ) )->toBe( $expected );
-} );
+//
+//use function WordPress\Streams\stream_str_replace;
```
adamziel, line 3: What's the issue with this test in PHP 7.4?

src/WordPress/Blueprints/BlueprintParser.php
adamziel: > The PHPStorm autolinter seems to be underconfigured.

You could configure it to use tabs instead of spaces, that would do the trick for now. And I agree, let's add linter in a subsequent PR.

composer.json
```diff
@@ -35,5 +42,8 @@
       "src/WordPress/Zip/functions.php",
       "src/WordPress/Streams/stream_str_replace.php"
     ]
+  },
+  "scripts": {
+    "phpcs": "phpcs --standard=PSR4"
```
adamziel: What would it take to use [WordPress coding standards](https://github.com/WordPress/WordPress-Coding-Standards) here?

src/WordPress/Blueprints/ResourceManager/Resolver/FilesystemResourceResolver.php
adamziel: > At this stage - will moving it back and then here again, fix the way it is displayed in this PR?

That is my understanding

src/WordPress/Blueprints/BlueprintParser.php
reimic: Sounds like a plan!

src/WordPress/Blueprints/ResourceManager/Resolver/FilesystemResourceResolver.php
reimic: Sadly, does not work as expected. From what I've read git should detect it as a directory rename, but it is a surprisingly complex git feature and might not deliver the trivially expected result. See: [directory-rename-detection](https://git-scm.com/docs/directory-rename-detection)

The scale of the issue is small - can we move past it?

src/WordPress/Blueprints/Runner/Step/CpStepRunner.php
reimic: Pretty much all of the things listed. Since the only thing it does is copy one specific file to a different location. IMHO that's constraining.

tests/Unit/stream_str_replace_Test.php
reimic: That is a Pest test. And we can't have Pest dependencies since they require a higher version of PHPUnit then the one we want to have. Also, keeping both Pest and PHPUnit, were it possible, would needlessly complicate things.

This applies to all commented tests. I'll rewrite them. I'd just prefer to do in separate PRs.

composer.json
reimic: Done. But linting and final standards adjustments will be done in a subsequent PR.

src/WordPress/Blueprints/Runner/Step/CpStepRunner.php
reimic: Let's move this discussion to #58 
Commit with cp-related changes and tests was reverted for this pr.

src/WordPress/Blueprints/Resources/Resolver/ResourceResolverCollection.php
```diff
@@ -0,0 +1,57 @@
...
+
+    public static function getResourceClass(): string
+    {
+        throw new \RuntimeException('Not implemented');
+    }
+
+    public function parseUrl(string $url): ?ResourceDefinitionInterface
+    {
+        foreach ($this->ResourceResolvers as $handler) {
+            $resourceFilePath = $handler->parseUrl($url);
```
adamziel: In here, `$resources` was renamed to `$resourceFilePath`. However, `$resourceFilePath` suggests a string while this is an object. Anything wrong with the name `$resource`?

src/WordPress/Blueprints/Resources/Resolver/ResourceResolverCollection.php
```diff
@@ -0,0 +1,57 @@
...
+            if ($resourceFilePath) {
+                return $resourceFilePath;
+            }
+        }
+
+        return null;
+    }
+
+    public function supports(ResourceDefinitionInterface $resourceDefinition): bool
```
adamziel: In here, `$resources` was renamed to `$resourceDefinition` which is fine but inconsistent with the rename in the `parseUrl` method. Let's choose one name and stick to it. I picked `$resource` because it's nice and short.

src/WordPress/Blueprints/Resources/Resolver/ResourceResolverCollection.php
```diff
@@ -0,0 +1,57 @@
...
+        foreach ($this->ResourceResolvers as $handler) {
+            if ($handler->supports($resourceDefinition)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public function stream(ResourceDefinitionInterface $resourceDefinition)
```
adamziel: Ditto

src/WordPress/Blueprints/Resources/Resolver/ResourceResolverInterface.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace WordPress\Blueprints\Resources\Resolver;
+
+use WordPress\Blueprints\Model\DataClass\ResourceDefinitionInterface;
+
+interface ResourceResolverInterface
+{
+    public function parseUrl(string $url): ?ResourceDefinitionInterface;
+
+    public function supports(ResourceDefinitionInterface $resourceDefinition): bool;
```
adamziel: This was also renamed from `$resource` to `$resourceDefinition`. That's a lot of changes for one PR. Let's stick to PHP 7.4 downgrade and move all the renames to another PR so this one can get merged faster.

adamziel:
Actually, what is the specific issue with the word `resource` in PHP 7.0? I just tried running the following code and it worked just fine:

```php
<?php

namespace Test\Resource;

$resource = 'test';

class Resource {}

function resource($resource) { echo "This is fine"; }

resource('a');
```

You can try it here: https://3v4l.org/nEQv0#v7.0.8

src/WordPress/Blueprints/ResourceManager/Resolver/FilesystemResourceResolver.php
adamziel: > The scale of the issue is small - can we move past it?

Okay, but reluctantly. This just means I need to manually put together a side by side diff to be able to review. It's just five files so I can do it, but let's ensure a proper rename in all future PRs.

adamziel:
This one is close. Let's conclude the discussion about the word Resource and resort back to using it if there are no issues with it, resolve the merge conflicts, and merge.

src/WordPress/Blueprints/ResourceManager/Resolver/FilesystemResourceResolver.php
reimic: Will do.

reimic: @adamziel this regards all the other open threads. I've responded to your inquiry and you gave a thumbs up to my response.

> Point 1: Resource -> ResourceManager -> Resources Yeah. It's a soft reserved word. [source](https://www.php.net/manual/en/reserved.other-reserved-words.php) It isn't a must, but it's a tiny ask and we can abide it. Nonetheless, I am changing the name to `Resources` as per your request.

You are right on 2 accounts - there is a naming inconsistency now, and this can be done elsewhere. So I'll revert the changes here.

reimic: > > The scale of the issue is small - can we move past it?
> 
> Okay, but reluctantly. This just means I need to manually put together a side by side diff to be able to review. It's just five files so I can do it, but let's ensure a proper rename in all future PRs.

Thanks for that. I've seen you made several commits, but I do not see any display / designation change. If you were able to get forward on that, could you guide me?

Ok. I have an idea what actually happened, correct me if I am wrong. As a mean to mitigate it being a delete/create not a move you moved it back, reviewed the changes within files, and then moved them again. If that is the case, that is really clever.

As a side note - I've noticed the underlying issue with PHPStorm's delete/create instead of move preference. The change history is broken. In hindsight, that should be obvious, but I assumed that a clear display of what changed in this PR is the broken part, whereas it just gets rid of the whole history.

src/WordPress/Blueprints/Resources/Resolver/ResourceResolverCollection.php
reimic: Reverted.

src/WordPress/Blueprints/Resources/Resolver/ResourceResolverCollection.php
reimic: Reverted.

src/WordPress/Blueprints/Resources/Resolver/ResourceResolverCollection.php
reimic: Reverted.

src/WordPress/Blueprints/Resources/Resolver/ResourceResolverInterface.php
reimic: Reverted.

reimic:
All done now. 😎

src/WordPress/Blueprints/ResourceManager/Resolver/FilesystemResourceResolver.php
adamziel: I've just tried to do `git mv` and failed, I think that's because an earlier commit in this branch deletes and recreates those files so they're already somehow 'disconnected' in Git history. Actually, I don't think git tracks any renames internally, [each commit is just a directory structure reflecting new files and dirs](https://www.alibabacloud.com/blog/a-detailed-explanation-of-the-underlying-data-structures-and-principles-of-git_597391). Would that mean the renames are inferred in runtime? 🤔 I don't know, but I'd love to find out in details. Either way, that's out of scope for this discussion.

adamziel:
@reimic This needs to be resolved before the PR can be merged:

CleanShot 2024-03-04 at 13 09 38@2x

src/WordPress/Blueprints/BlueprintParser.php
```diff
@@ -5,187 +5,206 @@
...
-			} else {
-				$subSchema = $subSchema->$key;
-			}
-		}
-
-		return $subSchema;
-	}
-
 
+// TODO Review class
```
adamziel, line 9: Let's use `@TODO:` for these to make them easier to find. Not a blocker here, just a note for the future. 

Also, this one isn't very clear. Why does this class need an additional review? What to look for?

src/WordPress/Blueprints/Resources/Resolver/InlineResourceResolver.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace WordPress\Blueprints\Resource\Resolver;
+namespace WordPress\Blueprints\Resources\Resolver;
```
adamziel, line 3: Nice, GitHub now shows these files side by side making them super easy to review

src/WordPress/Blueprints/Resources/ResourceMap.php
```diff
@@ -0,0 +1,60 @@
...
+
+    public function offsetGet($offset)
+    {
+        foreach ($this->pairs as $pair) {
+            if ($pair[0] === $offset) {
+                return $pair[1];
+            }
+        }
+
+        // TODO Evaluate waring: 'ext-json' is missing in composer.json
```
adamziel, line 36: Let's just add that extension to composer.json:

```json
{
   "require": {
        "ext-json": "*"
   }
}
```

adamziel:
I left a few nitpicks, but this PR looks pretty good otherwise. I'll go ahead and merge, let's address any remaining issues in a follow-up one.

src/WordPress/Blueprints/Resources/Resolver/InlineResourceResolver.php
reimic: Yeah. Surprisingly, the rebase helped.

src/WordPress/Blueprints/BlueprintParser.php
reimic: Ok, for the first part. Regarding the reason for marking it for review. It had a lot of commented out code, I wanted to signal that that could be addressed, but there was no reason to do that in this PR.

Issue #27, opened 2024-02-23
Set up a CI pipeline with GitHub's actions


reimic:
Automate the CI process with GitHub actions:

- [x] step running unit tests for PHP == 7.4, 
(as an intermediate step #25)
- [x] step running unit tests for PHP == 7.0,
(finally done with #86)
- [x] step with matrix unit tests for all versions > 7.0,
- [x] step with matrix unit tests for all versions > 7.4, 
(as an intermediate step #25)
- [ ] step for automatic linting, after #26
- [ ] step for coding standards check, after #26

PR #30, opened 2024-02-23
Set up a CI pipeline with PHPUnit tests

reimic:
🚧 Work in progress 🚧

Regarding #27 #25 

I've encountered a major problem, when trying to use PHP 8.2, so to move downgrading to a different pr. My goal is to rewrite Pest tests into PHPUnit test, but I can't make them work.

The test are failing with:
> PHP Fatal error:  Cannot acquire reference to $GLOBALS in D:\blueprints\vendor\phpunit\phpunit\src\Util\Configuration.php on line 543

I've read up on it, and it seems to be connected with the workflow setting up with PHP 8.2. But I am stuck.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,27 @@
+name: CI
+on:
+  push:
+    branches:
+      - trunk
+  pull_request:
+
+jobs:
+  test-unit:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.2'
+          tools: composer:v2
+          coverage: xdebug
+
+      - name: Install Dependencies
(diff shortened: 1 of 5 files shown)


reimic:
Full error log:

```
Run ./vendor/bin/phpunit
PHP Warning:  Private methods cannot be final as they are never overridden by other classes in /home/runner/work/blueprints/blueprints/vendor/phpunit/phpunit/src/Util/Configuration.php on line 176
PHP Fatal error:  Cannot acquire reference to $GLOBALS in /home/runner/work/blueprints/blueprints/vendor/phpunit/phpunit/src/Util/Configuration.php on line 543
Error: Process completed with exit code 255.
```

adamziel:
Oh boy, PHPUnit has its own PHP version compatibility matrix: https://phpunit.de/supported-versions.html

I looked up what does WordPress core do and [the composer.json](https://github.com/WordPress/wordpress-develop/blob/trunk/composer.json) file there seems to list the `"yoast/phpunit-polyfills": "^1.1.0"` package – perhaps that would help here? If not, you could ask on the #core channel on WP.org slack or #core-test

PR #57, opened 2024-03-02
Test Unzip Step

reimic:
### What does this PR do?

- Adds tests for the the UnzipStep.
- Refactors the UnzipStepRunner to require a progress tracker and handle `extraxt_to_path` in both relative and absolute form.
- Downgrades code where necessary to properly function for all PHP versions >= 7.4.
- Adds test resources. (For now in the `Step` directory - TBD where test resources will be stored.)
- Takes care of a TODO by canonicalizing a path in the `zip_extract_to` function.


src/WordPress/Blueprints/Model/DataClass/UnzipStep.php
```diff
@@ -2,60 +2,66 @@
...
+
+	/**
+	 * The file to extract
+	 *
+	 * @var string|ResourceDefinitionInterface
+	 */
 
-	/** @var string|ResourceDefinitionInterface */
-	public $zipFile;
+	// @TODO: Union types are PHP >= 8.0. Fix needed.
```
adamziel: Why add that type, then?

reimic: I come from a Java background. I guess, I still have to get used to fields being only partially typed...

adamziel: Gotcha! It's fine, in PHP types are optional, there could even be no annotation at all and it would still work.

src/WordPress/Blueprints/Model/DataClass/UnzipStep.php
```diff
@@ -2,60 +2,64 @@
...
+
 	public const DISCRIMINATOR = 'unzip';
 
 	/** @var Progress */
-	public $progress;
+	public Progress $progress;
 
 	/** @var bool */
-	public $continueOnError;
+	public bool $continue_on_error;
```
reimic: This is a WordPress coding standards-motivated change. But model data classes are derived from the scheme, so...   @adamziel does this mean we either change the coding standard or the scheme? As I understand, v2 is still in development...

reimic: Analyzing the CompiledStep class:

```
class CompiledStep {
  public function __construct (
    public StepDefinitionInterface $step,
    public StepRunnerInterface $runner,
    public Tracker $progressTracker,
  ) { }
}
```

What is the reasoning behind `$progress` and `$continue_on_error` being declared at individual step level? 

```
/** @var Progress */
public $progress;

/** @var bool */
public $continue_on_error;
```

For `$progress` I wonder if this is something that should be refactored as the `Tracker` object is now external to the `StepDefinition`? It is passed directly to the runner. Although both come from the `CompiledStep`. Why won't the run method accept a single argument implementing the `CompiledStep`? It this due to hermetization?

For `$continue_on_error`, I would imagine it could be a field declared at an abstract level. Although, complicating the inheritance structure for a single field does not seem wise. So maybe only a getter should be declared at interface level?

reimic:
image


This is very interesting and proves the usefulness of a matrix test step. My local environment is set to PHP 7.4 and when I run those tests they pass without issue. But they do fail here. This is critical info and I am glad it is there. (Someone really wise must've implemented it.)

src/WordPress/Blueprints/Model/DataClass/UnzipStep.php
adamziel: I would exclude the autogenerated code from PHPCS standards checks, only keep the PHP 7.0 compat checks around. Otherwise we'll keep running into annoying warnings that will require iterating on the code generator – there isn't much value in that.

adamziel: > What is the reasoning behind $progress and $continue_on_error being declared at individual step level?

`$progress` doesn't represent progress tracking but the progress part of the step declaration:

```js
{
	steps: [
		{
			step: 'defineWpConfigConsts',
			progress: {
				weight: 5,
				caption: 'Defining WordPress constants',
			}
		}
	]
}
```

About progress tracking:

BlueprintCompiler is what calculates the progress weight of each step and resource. If total progress is 100%, some steps may contribute 15% while others may contribute 27%. Each step reports its own progress, too. So from the perspective of a specific step, its own progress goes between 0% and 100%. It might report 10%, then 72%, then 100%. The top-level tracker reconciles that with that step's weight to calculate the total progress. Trackers are also hierarchical, so a top-level tracker may split into two sub-trackers, each representing 50% of the total progress, and each sub-tracker may split again. See the original TypeScript implementation for more context:

https://github.com/WordPress/wordpress-playground/blob/f9379b3fb26383a9ff874764ea1d27c95e994a56/packages/php-wasm/progress/src/lib/progress-tracker.ts#L72

---

About `$continue_on_error`, here's more context about its role:

* https://github.com/WordPress/wordpress-playground/issues/600
* https://github.com/WordPress/wordpress-playground/issues/604
* https://github.com/WordPress/wordpress-playground/pull/605

>  I would imagine it could be a field declared at an abstract level

Perhaps, I'm not that concerned about implementation details at this point. Let's leave it as it is for now and see what kind of abstractions emerge later on.

src/WordPress/Blueprints/Runner/Step/UnzipStepRunner.php
```diff
@@ -2,21 +2,33 @@
...
+	/**
+	 * Runs the Unzip Step
+	 *
+	 * @param UnzipStep    $input Step.
+	 * @param Tracker|null $progress Tracker.
+	 * @return void
+	 */
 	public function run(
 		UnzipStep $input,
 		Tracker $progress = null
```
adamziel: Let's require the tracker 

```suggestion
		Tracker $progress
```

reimic: Done.

reimic: Done.

reimic:
This PR became more about downgrading PHP than testing the Unzip Step. But that is a good thing, as downgrading the code becomes a priority. As a bonus - the matrix proved its usefulness.

adamziel:
src/WordPress/Blueprints/Model/DataClass/UnzipStep.php is autogenerated, let's revert all the changes applied to it.

src/WordPress/Blueprints/Runner/Step/UnzipStepRunner.php
```diff
@@ -2,21 +2,32 @@
...
+	 *
+	 * @param UnzipStep $input Step.
+	 * @param Tracker   $tracker Tracker.
+	 * @return void
+	 * @throws BlueprintException
+	 */
 	public function run(
 		UnzipStep $input,
-		Tracker $progress = null
+		Tracker $tracker
```
adamziel: Let's call it `progress_tracker` to make the purpose apparent. Alternatively, the `Tracker` class could be refactored to `ProgressTracker` – either of these work.

```suggestion
		Tracker $progress_tracker
```

src/WordPress/Blueprints/Runner/Step/UnzipStepRunner.php
```diff
@@ -2,21 +2,32 @@
...
-		$progress?->set( 10, 'Unzipping...' );
+		$tracker->set( 10, 'Unzipping...' );
 
-		// @TODO: Expose a generic helper method for this, e.g. $this->getExecutionContext()->resolvePath($input->extractToPath);
-		$toPath = $this->getRuntime()->getDocumentRoot() . '/' . $input->extractToPath;
-		zip_extract_to( $this->getResource( $input->zipFile ), $toPath );
+		$resolved_to_path = $this->getRuntime()->resolvePath( $input->extract_to_path );
+		try {
+			zip_extract_to( $this->getResource( $input->zip_file ), $resolved_to_path );
+		} catch ( \Throwable $exception ) {
```
adamziel: Not a blocker, but: I thought these try/catches would be handy in step runners, but in practice it's such a repetitive piece of boilerplate. It's too easy to forget about it or get it wrong in a single instance, plus it will be a pain to maintain. Let's just be optimistic about the execution result inside step runners and not rethrow BlueprintExceptions, and then catch and handle all exceptions in BlueprintRunner.

Would you mind starting a new issue to track this?

tests/Blueprints/Runner/Step/UnzipStepRunnerTest.php
```diff
@@ -0,0 +1,87 @@
...
+		$input = new UnzipStep();
+		$zip   = __DIR__ . '/test.zip';
+		$input->setZipFile( $zip );
+		$relative_path = 'dir';
+		$absolute_path = $this->runtime->resolvePath( $relative_path );
+		$input->setExtractToPath( $absolute_path );
+
+		$this->step->run( $input, new Tracker() );
+
+		$this->assertDirectoryExists( $absolute_path );
```
adamziel: Let's check if the directory actually contains the expected files and assert their content is as expected.

src/WordPress/Blueprints/Runner/Step/UnzipStepRunner.php
reimic: Sure. So this PR will not add the try-catch block for the Unzip Step Runner. I'll start a new issue: #65

tests/Blueprints/Runner/Step/UnzipStepRunnerTest.php
reimic: Done.

PR #58, opened 2024-03-02
Test Cp Step

reimic:
🚧 Work in progress 🚧

- TODO affirm it should overwrite newer files:
Filesystem's copy requires flag `overwriteNewerFiles` set to `true`
- TODO affirm it should copy file to directory
copy throws an exception when $resolvedToPath is a directory - linux would handle this:
When the program has one or more arguments of path names of files and following those an argument
of a path to a directory, then the program copies each source file to the destination directory,
creating any files not already existing.
- TODO affirm it should copy all files from one directory to another (also, ponder recursivity)
copy throws an exception when $resolvedToPath or $resolvedFromPath are directories - linux would
handle this:
When the program's arguments are the path names to two directories,
cp copies all files in the source directory to the destination directory,
creating any files or directories needed. This mode of operation requires an additional option flag,
typically r, to indicate the recursive copying of directories.
If the destination directory already exists, the source is copied into the destination,
while a new directory is created if the destination does not exist.
- TODO establish how $input->toPath should be handled
should copy to document root when null or when ""?

I've pondered it for a moment or two. And I think the right way forward here is to implement our own copy method.

My reasoning:

- As a user, I'd see a lot of value in using it the way unix specifies it. [source](https://en.wikipedia.org/wiki/Cp_(Unix)#:~:text=Meyering.%5B5%5D-,Operating%20modes,-%5Bedit%5D)
- Would work more akin to RmStep, handling both files and directories alike.
- We intend to have a limited amount of steps, so it's not like we'll have a ton of custom methods to maintain.
- We'll be less reliant on Filesystem, which we must downgrade to v3.x for PHP 7.0.

Currently the cp step copies only form one file path to another (both have to be files).


src/WordPress/Blueprints/Runner/Step/CpStepRunner.php
```diff
@@ -2,18 +2,53 @@
...
-		$success = copy( $input->fromPath, $input->toPath );
-		if ( ! $success ) {
-			throw new \Exception( "Failed to copy file from {$input->fromPath} to {$input->toPath}" );
-		}
-	}
+        $fileSystem = new Filesystem();
+
+        try {
+            // TODO affirm it should overwrite newer files
+            // Filesystem's copy requires flag `overwriteNewerFiles` set to `true`
```
adamziel, line 24: What are "newer" files? Is the timestamp compared?

src/WordPress/Blueprints/Runner/Step/CpStepRunner.php
```diff
@@ -2,18 +2,53 @@
...
+
+            // TODO affirm it should copy file to directory
+            // copy throws an exception when $resolvedToPath is a directory - linux would handle this:
```
adamziel, line 27: Yeah that would be useful.

src/WordPress/Blueprints/Runner/Step/CpStepRunner.php
```diff
@@ -2,18 +2,53 @@
...
+            // When the program has one or more arguments of path names of files and following those an argument
+            // of a path to a directory, then the program copies each source file to the destination directory,
+            // creating any files not already existing.
+
+            // TODO affirm it should copy all files from one directory to another (also, ponder recursivity)
```
adamziel, line 32: This would be highly useful as well.

src/WordPress/Blueprints/Runner/Step/CpStepRunner.php
```diff
@@ -2,18 +2,53 @@
...
+            // handle this:
+            // When the program's arguments are the path names to two directories,
+            // cp copies all files in the source directory to the destination directory,
+            // creating any files or directories needed. This mode of operation requires an additional option flag,
+            // typically r, to indicate the recursive copying of directories.
+            // If the destination directory already exists, the source is copied into the destination,
+            // while a new directory is created if the destination does not exist.
+
+            // TODO establish how $input->toPath should be handled
+            // should copy to document root when null or when ""?
```
adamziel, line 43: Let's assume all the paths are relative to the document root. `""` means "document root". `null` should be rejected during validation and never make it to this function.

src/WordPress/Blueprints/Runner/Step/CpStepRunner.php
reimic: Yes. Excerpt from the Filesystem docs:

> If the target file is older than the origin file, it's always overwritten. If the target file is newer, it is overwritten only when the $overwriteNewerFiles option is set to true.

I'd assume the steps should just do as they are told. Therefore, this suggestion.

adamziel: yeah, let's ignore the timestamps and just overwrite any pre-existing files

PR #63, opened 2024-03-05
Refactor method for better windows compatibility

reimic:
Refactor method `move_files_from_directory_to_directory` to enhance Windows compatibility.

Method will:
- use Filesystem's `rename` which handles various operating system better then the replaced core PHP method,
- canonicalize paths for better Windows compatibility.

### DIFF
--- a/src/WordPress/Blueprints/functions.php
+++ b/src/WordPress/Blueprints/functions.php
@@ -2,6 +2,8 @@
 
 namespace WordPress\Blueprints;
 
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Path;
 use Symfony\Component\Filesystem\Exception\IOException;
 use WordPress\Blueprints\Runtime\Runtime;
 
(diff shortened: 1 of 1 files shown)


src/WordPress/Blueprints/functions.php
```diff
@@ -42,16 +44,18 @@ function list_files( string $path, $omitDotFiles = false ): array {
...
 	foreach ( $files as $file ) {
 		if ( '.' === $file || '..' === $file ) {
 			continue;
 		}
-		$fromPath = $from . '/' . $file;
-		$toPath = $to . '/' . $file;
-		$success = rename( $fromPath, $toPath );
-		if ( ! $success ) {
-			throw new IOException( "Failed to move the file from {$fromPath} at {$toPath}" );
+		$fromPath = Path::canonicalize($from . '/' . $file);
```
adamziel, line 53: I'm not sure this is the right approach as it will turn `C:\Program Files\Stuff` into `C:/Program Files/Stuff`. It might be fine but I'm not sure if it will universally work with all Windows versions and paths, e.g. what about network shared directories such as `\\NetDirectory\SubPath`? 

That being said, perhaps it's good enough for now. Would you please start a new issue to get to the bottom of this eventually?

reimic: Follow up in issue: #66

PR #69, opened 2024-03-11
Reimplement JsonMapper 

reimic:
Reimplements parts of the Json Mapper to work in PHP == 7.0 and use less dependencies.

- Reimplements only the most crucial parts of the `jsonmapper/jsonmapper` dependency.
- Reworks `Middleware` and `Handlers` into simpler `Evaluators`. Instead of relying on magic methods for Handlers, a simple loop is run to go through all Evaluators.
- Reworks Evaluators to auto-configure just before mapping, not during initialization, as was the case prior to this PR. JsonMapper used to resolve all Handlers during initiation, later checking if the resolved values are still available when mapping methods were called. This caused some code to be run during container preparation. This was altered to configure at first `map_to_class` call, thus no mapper-related code is run before JSON input validation succeeds.
- Deletes the nonessential Evaluators convenience builder, still preserving elasticity for new Evaluators.
- Converts Enums to internal lists and strings. Enums were a separate dependency which required PHP > 7.0 and provided little utility. This relates to `ScalarTypes` and `Visibility` classes.
- Joins `PropertyMapper`, `ValueFactory`, `FactoryRegistry` and `ScalarCaster` into a single streamlined class.
- Internalizes some small utility methods to reduce class overengineering.
- Simplifies recursivity.
- Refactors all code to run with PHP == 7.0.
- Deletes all implementation parts that evaluated `void`, `mixed` , unions and etc. Since this mapper will only ever map to classes that are PHP 7.0 compliant.
- Adds tests.


src/WordPress/JsonMapper/Property/PropertyBuilder.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace WordPress\JsonMapper\Property;
+
+use WordPress\JsonMapper\ArrayInformation;
+
+class PropertyBuilder {
```
adamziel: Let's remove this and just use the Property class directly

src/WordPress/JsonMapper/Property/PropertyMap.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace WordPress\JsonMapper\Property;
+
+use ArrayIterator;
+use function array_key_exists;
+
+class PropertyMap implements \IteratorAggregate, \JsonSerializable {
```
adamziel: Not a blocker: Could this extend ArrayObject instead? Also, why does it need to be JsonSerializable? It seems like this class is all about a) having an object and not an array, and b) the `merge()` function. I wonder if a regular ArrayObject and a `merge_property_maps(array ...$maps)` function would also do the trick.

src/WordPress/JsonMapper/Evaluators/NamespaceResolver.php
```diff
@@ -0,0 +1,161 @@
+<?php
+
+namespace WordPress\JsonMapper\Evaluators;
+
+use PhpParser\NodeTraverser;
```
adamziel: This package is 250kb zipped. It's fine for this PR but let's explore switching to something lightweight like `token_get_all()` before shipping v1 of this library. Would you please start an issue to track that?

adamziel: This could come handy: https://gist.github.com/Zeronights/7b7d90fcf8d4daf9db0c

src/WordPress/JsonMapper/Evaluators/NamespaceResolver.php
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		return array_unique(
+			array_merge( self::getImportsForFileName( $filename ), self::getImports( $class->getParentClass() ) ),
+			SORT_REGULAR
+		);
+	}
+
+	/** @return Import[] */
+	private static function getImportsForFileName( string $filename ): array {
```
adamziel: Note for later: This might be problematic in eval()-like context where there's no file to parse. I don't think there's any other way to parse those `use` statements, though, so let's make it a known limitation of the `BlueprintMapper` documented in a doc block above that class.

src/WordPress/Blueprints/BlueprintParser.php
```diff
@@ -6,156 +6,76 @@
...
+
+	public function __construct(
+		BlueprintValidator $validator,
+		BlueprintMapper $mapper
+	) {
+		$this->validator = $validator;
+		$this->mapper    = $mapper;
+	}
+
+	function remove_utf8_bom( $text ) {
```
adamziel: Why is this needed?

src/WordPress/JsonMapper/Evaluators/PropertyMapper.php
```diff
@@ -0,0 +1,289 @@
...
+	 * @var JsonMapper
+	 */
+	private $mapper;
+
+	public function __construct(
+		JsonMapper $json_mapper
+	) {
+		$this->mapper = $json_mapper;
+		$this->add_factories_for_native_php_classes();
+		$this->add_factory_for_resources();
```
adamziel: This registers Blueprint-specific handlers in a `PropertyMapper` that looks generic. How easy would it be to move the registration of specific handlers to `BlueprintMapper`? And then, perhaps, the Mapper library would eventually become generic like `WordPress/JsonMapper`.

src/WordPress/JsonMapper/AnnotationMap.php
```diff
@@ -0,0 +1,64 @@
...
+
+	public function hasParam(string $paramName): bool
+	{
+		return array_key_exists($paramName, $this->params);
+	}
+
+	public function getParam(string $paramName): string
+	{
+		if (!$this->hasParam($paramName)) {
+			throw new \Exception("Annotation map doesnt contain param with name $paramName");
```
adamziel: Why not return null?

src/WordPress/JsonMapper/AnnotationMap.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace WordPress\JsonMapper;
+
+class AnnotationMap
```
adamziel: Is this for method annotations? Property annotations? Class annotations? It seems to hold information about `var`, `params`, and `return`. It would be nice if there was a documentation block to explain the semantics.

src/WordPress/JsonMapper/ArrayInformation.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace WordPress\JsonMapper;
+
+class ArrayInformation implements \JsonSerializable
```
adamziel: Does it need to be JsonSerializable?

src/WordPress/JsonMapper/ArrayInformation.php
```diff
@@ -0,0 +1,61 @@
...
+	{
+		return $this->isArray;
+	}
+
+	public function getDimensions(): int
+	{
+		return $this->dimensions;
+	}
+
+	public function isMultiDimensionalArray(): bool
```
adamziel: This method seems to be unused, let's remove it.

src/WordPress/JsonMapper/ArrayInformation.php
adamziel: Actually, this class seems to be just a container for two values: `isArray` and `dimensions`. Could it be refactored into either a simple array with two keys, or a simple class with two public properties and no methods?

src/WordPress/JsonMapper/Evaluators/NamespaceResolver.php
```diff
@@ -0,0 +1,161 @@
...
+		ObjectWrapper $object_wrapper,
+		PropertyMap $property_map,
+		JsonMapper $mapper
+	) {
+		foreach ( $this->fetchPropertyMapForObject( $object_wrapper, $property_map ) as $property ) {
+			$property_map->addProperty( $property );
+		}
+	}
+
+	private function fetchPropertyMapForObject( ObjectWrapper $object, PropertyMap $originalPropertyMap ): PropertyMap {
```
adamziel: nitpick: 
```suggestion
	private function computePropertyMapForObject( ObjectWrapper $object, PropertyMap $originalPropertyMap ): PropertyMap {
```

src/WordPress/JsonMapper/ObjectWrapper.php
```diff
@@ -0,0 +1,80 @@
...
+
+namespace WordPress\JsonMapper;
+
+use BadFunctionCallException;
+use ReflectionClass;
+use UnexpectedValueException;
+use function class_exists;
+use function is_null;
+
+class ObjectWrapper {
```
adamziel: Not a blocker for this PR: Is this class needed? Could we pass objects around instead?

src/WordPress/JsonMapper/JsonMapper.php
```diff
@@ -0,0 +1,42 @@
...
+		if ( null === $this->evaluators ) {
+			$this->evaluators = array(
+				new DocBlockAnnotations(),
+				new NamespaceResolver(),
+				new PropertyMapper( $this ), // PropertyMapper has to be the last evaluator.
+			);
+		}
+	}
+
+	public function map_to_class( \stdClass $json, string $class ) {
```
adamziel: I believe this is the generic term for this operation:

```suggestion
	public function hydrate( \stdClass $json, string $class ) {
```

src/WordPress/JsonMapper/JsonMapper.php
```diff
@@ -0,0 +1,42 @@
...
+use WordPress\JsonMapper\Evaluators\PropertyMapper;
+use WordPress\JsonMapper\Property\PropertyMap;
+
+class JsonMapper {
+	private $evaluators;
+
+	public function __construct() {}
+
+	private function configure_evaluators() {
+		if ( null === $this->evaluators ) {
```
adamziel: Let's do this in the constructor

src/WordPress/JsonMapper/JsonMapper.php
```diff
@@ -0,0 +1,42 @@
...
+			$this->evaluators = array(
+				new DocBlockAnnotations(),
+				new NamespaceResolver(),
+				new PropertyMapper( $this ), // PropertyMapper has to be the last evaluator.
```
adamziel: Let's document why it has to be the last one.

adamziel:
I left some notes, but things are mostly looking good. Yay! Thank you for you great work on this! I'd love to also see some more tests around:

* Arrays
* Union types
* Nullable and missing properties

For example, the following Blueprint should be correctly mapped even though the array contains two different data types:

```json
{
	"plugins": [
		"https://downloads.wordpress.org/plugin/wordpress-importer.zip",
		{ "resource": "url", "url": "https://mysite.com" }
        ]
}
```

However, this Blueprint should be rejected since numbers are not valid as a plugin definition:

```json
{
	"plugins": [
		"https://downloads.wordpress.org/plugin/wordpress-importer.zip",
		123
        ]
}
```

src/WordPress/Blueprints/Model/BlueprintBuilder.php
```diff
@@ -123,19 +128,34 @@ public function withFiles( $files ) {
...
-	public function downloadWordPress( string|ResourceDefinitionInterface $wpZip = null ) {
-		$this->prependStep( ( new DownloadWordPressStep() )
+	public function remove( $path ) {
+		return $this->addStep(
+			( new RmStep() )
+			->setPath( $path )
+		);
+	}
+
+	public function makeDirectory( $path ) {
```
adamziel: Let's remove these methods for now.

First, this is additional code to maintain and keep in sync with the MkdirStep signature. 

Second, it doesn't expose all the MkdirStep features, e.g. you can't specify progress information here.

Third, and this is the most important one, the Builder API needs a deeper design dive to make sure the method names communicate the semantics well. The name `makeDirectory` doesn't make the consequences apparent. Does it add a new step? Does it create a directory immediately? If it's called together with some shorthand methods, what will be the order of execution? Also, if a Blueprint plugin registers more steps, there won't be shorthand methods for them – is that okay or not? These questions needs answers before v1 is released.

src/WordPress/JsonMapper/JsonMapper.php
reimic: From PR description:

> Reworks Evaluators to auto-configure just before mapping, not during initialization, as was the case prior to this PR. JsonMapper used to resolve all Handlers during initiation, later checking if the resolved values are still available when mapping methods were called. This caused some code to be run during container preparation. This was altered to configure at first map_to_class call, thus no mapper-related code is run before JSON input validation succeeds.

src/WordPress/JsonMapper/Evaluators/NamespaceResolver.php
reimic: Great point. Will monitor in issue #70

reimic:
> I left some notes, but things are mostly looking good. Yay! Thank you for you great work on this! I'd love to also see some more tests around:
> ...
> Union types

I do not see how union types play into this - they are not a PHP 7.0 thing. I would imagine we should just not use them anywhere in the lib.

reimic:
> I left some notes, but things are mostly looking good. Yay! Thank you for you great work on this! I'd love to also see some more tests around:
> Arrays
> Union types
> Nullable and missing properties

Most definitely. I do have quite a lot of tests in mind. How would you like to progress from here: test the mapper now or merge it, finish other downgrade-related matters and than circle back? 

I'd recommend the latter. I think, and this is also reflected in your comments, there are still some things we can iterate upon. A testing-focused PR done in a week or so would be a great occasion to go through everything and make some improvements with "fresh" eyes.

src/WordPress/JsonMapper/JsonMapper.php
adamziel: I'm not sure I follow. So merely setting up an array of evaluators conflicts with the container setup? That's unexpected, it sounds like lower level structures depend on higher level structures here.

adamziel:
> I do not see how union types play into this - they are not a PHP 7.0 thing. I would imagine we should just not use them anywhere in the lib.

Types are expressed either as type annotations or as comments. Union types are all over Blueprints, e.g. `ResourceInterface|string`, they're just stored in comments since PHP 7.0 doesn't support them. Ditto for typed collections like `string[]`.

src/WordPress/Blueprints/Model/BlueprintBuilder.php
reimic: I've reviewed the Blueprint Builder. 
It's a great idea, but still in its infancy. Many methods are implemented in a way that does not fully expose their features. Yet, using the Blueprint Builder for test data is a boon. 

If this Builder is still so floaty, maybe we can let those methods be and create a full issue to ideate on it and on all considered mapping strategies. 

I had many of the questions you asked when I was writing tests that coincidentally tested mapping to the `InstallPluginStep` and realized it has a shorthand. Blueprint Builder allows only for the shorthand. The JSON schema allows for both shorthand and steps. I have found this really confusing. First, you can't ideally reflect one mapping strategy to another i.e. there is no way to Build a Blueprint so that it is identical to one achieved via JSON. I don't exactly know how I feel about it, but I know it is not stellar.

See this test:
https://github.com/WordPress/blueprints/blob/7415d32f5886c2c6f3f4e8ea70f800e2874d6644/tests/JsonMapper/JsonMapperTest.php#L82-L111

It fails. The result, as one can expect, is that the mapper returns an object with some plugins in the shorthand plugin list, and some as steps. This discrepancy is then corrected in the `BlueprintCompiler` with [expandShorthandsIntoSteps](https://github.com/WordPress/blueprints/blob/f2ddfa54a97220249201535f1d37726c416f915d/src/WordPress/Blueprints/Compile/BlueprintCompiler.php#L32). Yet this feels late. Ideally all 4 mapping strategies: `from Blueprint Builder`, `from Blueprint`, `from JSON` and `from JSON string` should yield the same object.

reimic: Does this need a separate issue or is it covered by the general scheme ponderings?

Ah. Important remark - I love the idea of shorthands, just this one is not sufficiently fleshed out yet.

adamziel:
Let's test the mapper now to make sure it actually works. It's a central, critical part of the system. Breaking it would generate more work with debugging, reverting this or etc.

src/WordPress/JsonMapper/JsonMapper.php
reimic: There is no conflict. It runs code it might not need. It can happen at this stage. With the magic method handlers it made some unexpected things, but now with the code streamlined it should be less of a problem.

But still - this is run always once, and I think if there is a chance JSON validation will fail and this code won't be needed there is no value in instantiating the classes at container creation.

src/WordPress/Blueprints/Model/BlueprintBuilder.php
adamziel: > If this Builder is still so floaty, maybe we can let those methods be and create a full issue to ideate on it and on all considered mapping strategies.

No, that will generate more noise to process later on. Let's just keep it as it is for now.

src/WordPress/JsonMapper/Evaluators/PropertyMapper.php
reimic: Easy enough. So easy in fact that it can be done when it is needed. For now, with no other application of the Json Mapper in sight - the code rolls quite nicely as it is, no additional variables need to be passed when instantiating a specific factory, as all are instantiated as private methods in this class.

In fact, with what the mapper is now, I wonder that purpose do those two classes have: `BlueprintMapper` and `JsonMapper`. With a generic mapper - sure, they should be separate. But for now?

src/WordPress/JsonMapper/JsonMapper.php
adamziel: > But still - this is run always once, and I think if there is a chance JSON validation will fail and this code won't be needed there is no value in instantiating the classes at container creation.

It's a microoptimization that, most likely, no benchmark would be able to detect. It's not worth additional complexity, let's simplify.

src/WordPress/JsonMapper/Evaluators/PropertyMapper.php
adamziel: With this structure, maintaining these handlers will require diving into internals of the JSON mapper. It's counter-intuitive to look for implementation-specific handlers in a structure of otherwise generic classes. If it's as easy, let's just do it. Discussing, creating tracking issues, looking for these factories in codebase will likely take more time than just making this change.

src/WordPress/Blueprints/BlueprintParser.php
reimic: I'll remove `remove_utf8_bom()` for now.

However there are 2 aspects here. I think each deserves an issue of its own.

1. I've encountered multiple problems with `json_decode()`. It was very sensitive to a bunch of things related to Windows end line characters and general formatting. This should be looked at as a semantically and syntactically valid JSON might be provided, but some miniscule formatting issues will cause it to fail at decoding. Should be reviewed before v1.
2. Despite the above. It will always be possible that the JSON will me in some shape or form syntactically malformed. In such cases `json_decode()` returns null and `json_last_error()` can be called to get more information on what actually transpired. I've added the most basic `Malformed JSON.` message, just to signify the issue was syntactic, not semantic in nature. But I know from experience this is one of the most frustrating parts of using a new API. If we truly aspire to deliver a premium way of setting up/developing WordPress, providing a robust troubleshooting experience is paramount. That said, this is not a v1 deployment-blocking issue.

src/WordPress/JsonMapper/Evaluators/PropertyMapper.php
reimic: Done. [See this commit.](https://github.com/WordPress/blueprints/pull/69/commits/da346af421e0c42c25d190799e03f60bee3456cb)

src/WordPress/JsonMapper/JsonMapper.php
reimic: Was also handles with https://github.com/WordPress/blueprints/pull/69#discussion_r1521329385

Optimization was not a real concern of mine. The issue was to ease debugging. When I was learning what the code does my breakpoints for JsonMapper's Handlers' internals would fire even before a single map call was made. This was partially due to use of `__invoke`. It confused me immensely, so at one point I deemed it counter-intuitive. 

With the current, simpler implementation this is a non-issue, so I see why you would look at performance to understand my reason for the change.

src/WordPress/JsonMapper/ArrayInformation.php
reimic: I agree. I'll see if it is truly needed at all and how to implement it in the simplest way possible.

src/WordPress/JsonMapper/ArrayInformation.php
reimic: Sure. Was made obsolete by simplifying the recursive methods in the Property Mapper.

src/WordPress/JsonMapper/Evaluators/NamespaceResolver.php
reimic: Sure. Changed to `compute_property_map` since it takes an object and it is the only computing method there.

src/WordPress/JsonMapper/ObjectWrapper.php
reimic: It is not useless, that is for sure. I'd happily get rid of it while writing tests to assure no features were lost in the process. This might happen in this PR, with the core tests. Or later, with better test coverage. We'll see.

src/WordPress/JsonMapper/AnnotationMap.php
reimic: Why indeed... This was the convention the Json Mapper had. I chose not to meddle with it. But the whole `AnnotationMap` might just be an internal array in the `DocBlockAnnotations` Evaluator.

src/WordPress/Blueprints/BlueprintParser.php
adamziel: Would you please start a new issue with the problem description and reproduction steps?

src/WordPress/Blueprints/BlueprintMapper.php
```diff
@@ -4,16 +4,22 @@
...
-	protected $mapper;
+	private $mapper;
+
+	private $custom_factories = array();
 
 	public function __construct() {
 		$this->mapper = new JsonMapper();
+		$this->prepare_factory_for_resources();
+		$this->prepare_factory_for_steps();
+		$this->mapper->configure_evaluators( $this->custom_factories );
```
adamziel: Nitpick, how about an API like this?

```suggestion
		$this->mapper = new JsonMapper(array(
			$this->create_resource_mapper(),
			$this->create_steps_mapper()
		));

```

adamziel:
This is looking better and better ❤️

src/WordPress/JsonMapper/AnnotationMap.php
reimic: For our use cases `AnnotationMap` could've been dropped entirely. So that is was has happened now.

adamziel:
@reimic the PR description covers a delta between the JsonMapper library and this PR. It would be more useful if it instead discussed the problem being solved, gave a short overview of the solution, and provided testing instructions.

reimic:
> @reimic the PR description covers a delta between the JsonMapper library and this PR. It would be more useful if it instead discussed the problem being solved, gave a short overview of the solution, and provided testing instructions.

Fair point. I'll do one, after all threads are resolved.

src/WordPress/JsonMapper/Property/PropertyBuilder.php
reimic: Done.

src/WordPress/JsonMapper/Property/PropertyMap.php
reimic: Was turned into a simple array of Properties.

src/WordPress/JsonMapper/Evaluators/NamespaceResolver.php
reimic: I would like to take a look at those imports in a subsequent PR. For now - what type o message do you propose?

src/WordPress/JsonMapper/ArrayInformation.php
reimic: Gone. Gone like the wind.

src/WordPress/JsonMapper/JsonMapper.php
reimic: Done.

src/WordPress/JsonMapper/JsonMapper.php
reimic: Issue resolved entirely with architecture changes.

src/WordPress/Blueprints/Model/BlueprintBuilder.php
reimic: Done.

src/WordPress/Blueprints/BlueprintMapper.php
reimic: Done.

src/WordPress/Blueprints/Model/DataClass/Blueprint.php
```diff
@@ -18,7 +20,7 @@ class Blueprint
 
 	/**
 	 * Slot for runtime–specific options, schema must be provided by the runtime.
-	 * @var \ArrayObject
+	 * @var ArrayObject
```
adamziel: Is this merely a style change, or is there a technical reason for this?

src/WordPress/JsonMapper/Evaluators/NamespaceResolver.php
reimic: Gone. Imports do not seem relevant for the mapper at this point in time.

src/WordPress/Blueprints/Model/DataClass/Blueprint.php
reimic: Was a test to see if the mapper handles imported and prefixed classes the same way. It did not, now it does.

src/WordPress/Blueprints/Model/DataClass/ActivatePluginStep.php
```diff
@@ -19,7 +19,7 @@ class ActivatePluginStep implements StepDefinitionInterface
 	 * Plugin slug, like 'gutenberg' or 'hello-dolly'.
 	 * @var string
 	 */
-	public $slug;
+	public $slug = null;
```
adamziel: `null` is the default value for all uninitialized field and is okay to skip from the declaration.

src/WordPress/JsonMapper/JsonMapper.php
```diff
@@ -0,0 +1,362 @@
...
+	 */
+	public function __construct( array $custom_factories = array() ) {
+		$this->add_factories_for_native_php_classes();
+		$this->add_custom_factories( $custom_factories );
+	}
+
+	/**
+	 * Creates an instance of the given class and populates it with data from the given JSON object.
+	 *
+	 * This method first checks if a factory exists for the given class. If a factory exists, it uses the factory
```
adamziel: Is it GPT generated? That docstring is of negative value – it doesn't add any context on top of the implementation and it also has to kept in sync with the code changes. Let's remove all the docstring lines that are just literal english transcriptions of specific code lines below. A useful docstring says something about the method or usage that isn't immediately obvious otherwise.

src/WordPress/JsonMapper/PropertyParser.php
```diff
@@ -0,0 +1,134 @@
...
+		return 'private';
+	}
+
+	/**
+	 * Returns an array of property types parsed from the provided DocBlock. Filters out 'null'
+	 * and removes Global Namespace Prefixes.
+	 *
+	 * Examples:
+	 *
+	 * For types: PointlessInterface[]|null will return: array('PointlessInterface[]')
```
adamziel:
```suggestion
	 * For types: AnInterface[]|null will return: array('AnInterface[]')
```

reimic:
This comes from fixing the bug in the model generator. Where it either can
set an inconsequential null in some places or not set defauls "" for
string, false for bool, or 0 for int.

reimic:
Sure. Why not.

src/WordPress/JsonMapper/PropertyParser.php
reimic: Sure. Why not.

src/WordPress/Blueprints/Model/DataClass/ActivatePluginStep.php
reimic: This comes from fixing the bug in the model generator. Where it either can
set an inconsequential null in some places or not set defaults "" for
string, false for bool, or 0 for int.

adamziel: I'm not sure what the issue was but I removed these nulls without affecting `""`, `false`, and `0` values in 8ddb3c8b783edc4a03f686b735d0cf8b3dca6f57.

adamziel (approved):
Let's get this in! @reimic would you please update the description for any future readers as discussed earlier?

src/WordPress/Blueprints/Model/DataClass/ActivatePluginStep.php
reimic: Sure. I get where this is coming from.

Previously the code looked like this:
``` 
if ( $schema->getDefault() ) {
   $property->setValue( $schema->getDefault() );
}
```
`$schema->getDefault()` could return a non-falsy value where no issue occurred or one of these: null, 0, "" and false.

I've removed the check entirely and assigned all values, null included. You've enhanced it to make a strict check, allowing falsy values other than null to be assigned.
``` 
if ( null !== $schema->getDefault() ) {
   $property->setValue( $schema->getDefault() );
}
```

src/WordPress/JsonMapper/JsonMapper.php
reimic: It was. But I've edited it heavily. I guess I was swayed by your inclination to make this library as beginner-friendly (any skillset dev friendly) as possible and I overdid it here. Thanks for your remarks. I've analyzed the DocBlocks you've committed, so it will be easier to hit closer to the target next time. Thanks for that!

PR #77, opened 2024-03-19
Transpile to PHP 7.0 with Rector and manual replacements

adamziel:
## PHP 7.0 Compatibility

Makes the Blueprints library syntax–compatible with PHP >= 7.0 with automated transpilation and manual adjustments.

The automated transpilation is done with [rector](https://github.com/rectorphp/rector),
which downgrades features specific to PHP 7.2 and later features to PHP 7.1.

The manual part requires using regexps and editing the files by hand.

### Automated part

To transpile the code to PHP 7.1, run:

```bash
# Install rector:
composer require rector/rector --ignore-platform-req=php

# Transpile:
php vendor/bin/rector process src
```

Unfortunately, Rector does not support downgrading to PHP 7.0 yet, so we need to do the
last stretch manually.

### Manual part

Rector will downgrade PHP code to PHP 7.1 but not further. We need PHP 7.0 compat
so here's a few additional regexps to run. Regexps are not, of course, reliable in
the general case, but they seem to do the trick here.

List of manual replacements

* `: \?[a-zA-Z_0-9]+` -> (empty string) to remove the unsupported return type
  from `function(): ?SchemaResolver {}` -> `function() {}`.
* `: iterable` to fix `Fatal error: Generators may only declare a return type of Generator, Iterator or Traversable`.
* `\?[a-zA-Z_0-9]+ \$` -> `$` to remove the unsupported nullable type from function signatures,
  e.g. `function(?Schema $schema){}` -> `function($schema){}`.
* `(protected|public|private) const` -> `const` as const visibility is not supported in PHP 7.0.
* `: void` -> `` as `void` return type is unsupported in PHP 7.0.
* `[$ns, $name] = $this->parseName($name);` -> `list($ns, $name) = $this->parseName($name);`
* `foreach ($data as [$cp, $chars]) {` -> `foreach ($data as list($cp, $chars)) {`
* Find or write Rector rules for downgrading to PHP 7.0

### DIFF
--- a/src/WordPress/AsyncHttp/Client.php
+++ b/src/WordPress/AsyncHttp/Client.php
@@ -89,7 +89,7 @@ use function WordPress\Streams\streams_send_http_requests;
  */
 class Client {
 	protected $concurrency = 10;
-	protected Map $requests;
+	protected $requests;
 	protected $onProgress;
 	protected $queue_needs_processing = false;
 
@@ -150,7 +150,7 @@ class Client {
 	 *
 	 * @return resource
 	 */
-	public function get_stream( Request $request ) {
+	public function get_stream( $request ) {
 		if ( ! isset( $this->requests[ $request ] ) ) {
 			$this->enqueue_request( $request );
 		}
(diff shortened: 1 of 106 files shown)


adamziel:
There's plenty of formatting issues in this PR, let's reformat the entire repo in a follow-up one.

PR #82, opened 2024-03-19
Add e2e tests

reimic:
🚧 Work in progress 🚧

What does this PR do?
- adds e2e tests

todo:
- set phpunit version to 6.5
- fix all unit tests to match phpunit == 6.5
- fix issues uncovered in tests
- examine why a test failing validation is displayed as passing
- fix assertions

### DIFF
--- a/src/WordPress/AsyncHttp/Client.php
+++ b/src/WordPress/AsyncHttp/Client.php
@@ -192,7 +192,7 @@ class Client {
 
 		foreach ( $streams as $k => $stream ) {
 			$request                            = $enqueued[ $k ];
-			$total                              = $response_headers[ $k ]['headers']['content-length'];
+			$total                              = $response_headers[ $k ]['headers']['content-length'] ?? null;
 			$this->requests[ $request ]->state  = RequestInfo::STATE_STREAMING;
 			$this->requests[ $request ]->stream = stream_monitor_progress(
 				$stream,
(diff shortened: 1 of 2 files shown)


reimic:
@adamziel it would be cool if you could take a look at supposedly passing tests for PHP == 8.3 (hint - they do not pass):

The CI test `Running json blueprint with steps` fails at step `setSiteOptions`, which is quite a surprise. Locally it fails at `defineWpConfigConsts` which indicates a Windows compatibility issue, again.

The CI test `Running json blueprint with word press version` fails only because I did not properly prepare what should be asserted. But locally I get an error for step `downloadWordPress`. Again, most likely a Windows problem... 

For php blueprints - all fail at validation. I'll make an issue for that.... and I'll add another one to enhance the quality of errors thrown.

```
Array
(
    [/constants] => Array
        (
            [0] => The data (array) must match the type: object
        )

)
```

That's not the most informative thing to get from a validator. We can do better.

adamziel:
Let's chat in https://github.com/WordPress/blueprints-library/issues/83

tests/e2e/ShorthandsTest.php
```diff
@@ -0,0 +1,107 @@
...
+			self::assertEquals( $result->step, $expected_steps[$key] );
+		}
+	}
+
+	/**
+	 * Data provider for {@see self::testRunningBlueprintWithWordPressVersion()}.
+	 *
+	 * @return array
+	 */
+	public function blueprint_with_word_press_version() {
```
adamziel: Nitpick: it’s WordPress, not word_press. Let’s either stick to the correct word or say just `wp`

Issue #65, opened 2024-03-06
Exception handling when running steps


reimic:
Most methods executed by step runners can raise a exceptions. Until now this was handled within those methods, but it has proven itself to be quite cumbersome. It's quite easy to forget to properly wrap a method invocation with a try-catch and thus allowing for an issue in a single step to fail a whole blueprint. Also - that's quite a boilerplate-heavy solution.

As a potential solution - exceptions would be handled in the orchestrating `BlueprintRunner.php`, thus "globally" protecting the execution flow, with no need for extra boilerplate and no risk of exception-leakage.

PR #92, opened 2024-03-26
Refactor MkdirStepRunner and test it

reimic:
### What does this PR do?

`MkdirStepRunner` will now:
- handle both relative and absolute paths
- throw an exception if the directory it wants to create already exists
- throw an original Symfony Filesystem `IOException` not rewrapping it into a `BlueprintException`, as agreed in #65

### What problem does it fix?
- until now the runner would only handle absolute paths
- has thrown an exception in a bad way
- we use this step in the `examples` directory, and since it only accepted absolute paths and we have provided it a relative path the example was flawed

### How to test if it works?
- this PR includes a full suite of tests for this runner

### DIFF
--- a/src/WordPress/Blueprints/Runner/Step/MkdirStepRunner.php
+++ b/src/WordPress/Blueprints/Runner/Step/MkdirStepRunner.php
@@ -2,19 +2,22 @@
 
 namespace WordPress\Blueprints\Runner\Step;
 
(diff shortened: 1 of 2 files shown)


adamziel:
LGTM! I left two nitpicks. Happy to merge once they're fixed!

reimic:
Doneski, boss.

Issue #73, opened 2024-03-17
Zip Slip Vulnerability


Lucisu:
As we discussed today, there's a ZIP Slip vulnerability below here. The filename should be sanitized to prevent directory traversal:

https://github.com/WordPress/blueprints-library/blob/f2ddfa54a97220249201535f1d37726c416f915d/src/WordPress/Zip/functions.php#L17C51-L17C62

Also, maybe it's good to check for Symlinks and how they are being treated nowadays ;)

PR #93, opened 2024-03-26
Fix zip slip vulnerability, other zip issues and tests

reimic:
🚧 Work in progress 🚧

### What does this PR do?
- `zip_extract_to` method will now throw an exception if it encounters files with a relative path #73  ( I recommend throwing an exception instead of skipping dangerous files in zips. I was swayed by the argument that letting only some files through breaks [user-data integrity](https://github.com/OWASP/ASVS/issues/1424#issuecomment-1369726583).) 

### What problem does it fix?
- there was a [vulnerability](https://security.snyk.io/research/zip-slip-vulnerability) in the `zip_extract_to` method

### How to test if it works?
- this PR explicitly tests that the zip slip / zip symlink vulnerability was fixed


brandonpayton:
Hi @reimic, thank you for this PR. As I was reading, I found it a bit difficult to review due to multiple purposes in the changes:

> ### What problem does it fix?
> * there was a [vulnerability](https://security.snyk.io/research/zip-slip-vulnerability) in the `zip_extract_to` method
> * zip functions caused an error when the code was run with PHP == 7.0
> * tests were slightly off and had to be updated

Would you be willing to break this into multiple single-purpose PRs so we can focus more clearly on each issue?

reimic:
Sure, @brandonpayton - let's start here: #94

reimic:
Annnd, @brandonpayton - then progress to here: #95

tests/unit/zip/ZipFunctionsTest.php
```diff
@@ -3,28 +3,17 @@
...
-		$zip->close();
-
-		zip_extract_to( fopen( $filename, 'rb' ), dirname( $filename ) );
-
-		$slipped_dir = Path::canonicalize(__DIR__ . "../../../../../../../../tmp");
-		self::assertDirectoryDoesNotExist( $slipped_dir );
-
-		$filesystem->remove( dirname( $filename ) );
+		$slipped_file = Path::canonicalize(__DIR__ . "../../../../../../../../tmp/zip-slip-test.txt");
+		self::assertFileDoesNotExist( $slipped_file );
```
adamziel: This checks for a single, very specific path. Why not test for a single, simple case like `../tmp-zip-slip-test.txt`? And then confirm where the file was actually created – if anywhere?

adamziel: Also let's test for a path starting with `/`

adamziel: Also let's test for zipped symlinks

Issue #104, opened 2024-06-04
HTTP client doesn't handle responses with Transfer-Encoding: chunked


maypaw:
In the process of working on #82 (E2E tests) with @reimic we found an issue with the way how the HTTP client handles chunked streams. 

Currently, the response is written with the chunk markers, as if they were normal characters. This corrupts the wp-cli.phar. Handling Transfer-Encoding: chunked is expected behaviour of HTTP/1.1 protocol. 

At the moment we have two ideas how to handle this issue:
 1. Change the used HTTP protocol version to HTTP/1.0, which by default doesn't support chunks.
 Currently the function `stream_http_prepare_request_bytes` in `async_http_streams.php` prepares a request in 1.1: 
 ```
$request = <<<REQUEST
GET $path HTTP/1.1 // We could change this to: GET $path HTTP/1.0
[...]
REQUEST;
```
We've tested this approach and it works. The downside to this would be probable performance loss, since HTTP/1.1 is supposedly more efficient. Yet neither I nor @reimic are capable of assessing the influence of such change.

 2. Alternatively, the streams could be filtered with the PHP dechunk filter when performing read or write operations. We have also tested that this can work. Yet this is somewhat problematic:
We tried appending the filter after the stream creation, in function `streams_http_open_nonblocking` in `async_http_streams.php` and several other places. Unfortunately, due to usage of `@stream_select` in continuous reading and writing the response, appending stream filter while performing those operations is impossible (reference: https://github.com/reactphp/event-loop/issues/220). 
The first place, where it worked as expected, was at the very last possible moment, i.e. while copying the already closed stream in `run` method of `WriteFileStepRunner.php`.  
 ```
public function run(
		$input,
		$progress = null
	) { 
         [...]
		$fp2 = fopen( $path, 'w' );
         [...]
         stream_filter_append($fp2,"dechunk",STREAM_FILTER_WRITE); // dechunking the stream
		stream_copy_to_stream( $this->getResource( $input->data ), $fp2 );
		fclose( $fp2 );
}
```
This doesn't seem like the right place to do this though, as dechunking should probably be done when handling the response in the client.

@adamziel what do you think?

adamziel:
Great issue! Let's use HTTP 1 until we have chunked encoding. Chunked encoding would have to be likely implemented here

```php
function streams_http_response_await_bytes( $streams, $length, $timeout_microseconds = 5000000 ) {
	$read = $streams;
	if ( count( $read ) === 0 ) {
[...]

Issue #60, opened 2024-03-05
Windows compatibility


adamziel:
Let's test the Blueprints library on Windows and make sure it runs smoothly there

Done is:

* There's a fairly comprehensive test suite in the repo (definition of _fairly comprehensive_ TBD)
* It passes on both Linux and Windows

cc @reimic

reimic:
Updated list of sub-issues on this topic:

- [ ] #66 
- [x] #106 (fixes a request parsing issue) 
- [x] #103 (fixes missing zip extension in the CI test pipeline)

PR #106, opened 2024-06-08
Fix issues in HTTP client

maypaw:
### Summary
This pull request introduces fixes to the HTTP client as described in #104.

### Major Changes:
- temporary change of protocol HTTP/1.1 -> HTTP/1.0 - HTTP/1.0 doesn't support Transfer-Encoding: chunked and thus temporarily solves the dechunking problem, until some more robust solution is implemented,
- additional null-check for Content-Length header, when assigning total bytes size in HTTP client, which otherwise resulted in an `PHP Notice:  Undefined index: content-length`
- removal of excessive line-ending replacement - the explicit replacement from "\n" to "\r\n" broke sending requests on Windows.

### How to Test
After checking out to the feature branch, run `blueprint_compiling_simple_progress.php` script. The script passes all steps connected with downloading and installing WordPress (but still does not complete). 

### Notes
- the `blueprint_compiling_simple_progress.php` script fails on my machine at step DefineWpConfigConsts, at the moment I'm not sure why, but I'll create a separate issue describing the problem
- the change of protocol is only temporary and will allow further work on #82 @reimic

### DIFF
--- a/src/WordPress/AsyncHttp/async_http_streams.php
+++ b/src/WordPress/AsyncHttp/async_http_streams.php
@@ -219,18 +219,17 @@ function stream_http_prepare_request_bytes( $url ) {
 	$parts   = parse_url( $url );
 	$host    = $parts['host'];
 	$path    = $parts['path'] . ( isset( $parts['query'] ) ? '?' . $parts['query'] : '' );
-	$request = <<<REQUEST
-GET $path HTTP/1.0
-Host: $host
-User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/93.0.4577.82 Safari/537.36
-Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9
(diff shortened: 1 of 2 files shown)


reimic:
Splendid! Welcome aboard, Maya. 

Check this out @adamziel 

For a bit of housekeeping - this is also relevant to #60

src/WordPress/AsyncHttp/async_http_streams.php
```diff
@@ -230,7 +230,7 @@ function stream_http_prepare_request_bytes( $url ) {
 
 	// @TODO: Add support for Accept-Encoding: gzip
 
-	return str_replace( "\n", "\r\n", $request ) . "\r\n\r\n";
+	return $request . "\r\n\r\n";
```
adamziel: It seems like we'd still get inconsistent newline style depending on where this file was last saved and where is it used. Let's declare each line of the request separately and concatenate them with the \r\n sequence so the result is always the same.

adamziel (changes requested):

src/WordPress/AsyncHttp/async_http_streams.php
maypaw: The newest change introduces single newline style \r\n.

PR #112, opened 2024-07-13
AsyncHttp: Support transfer-encoding

adamziel:
An exploration to see what kind of internal refactor might be required to support `Transfer-Encoding: chunked` and `Content-encoding: gzip`. Related to #104. cc @MayPaw – I took a stab on this one as I worry it might soon become a blocker for the data liberation work like downloading asset files, see https://github.com/WordPress/data-liberation/discussions/74.

There's no test plan yet, other than running `php http_api.php`.


adamziel:
Superseded by https://github.com/WordPress/blueprints-library/pull/113

PR #115, opened 2024-10-25
Support composer test command

brandonpayton:
This is a simple PR that allows tests to be run with `composer test`, as one might guess they could be run.

This also fixes an unrelated typo.


zaerl (approved):
I see some unit tests complaining about changes in `PHPUnit\TextUI\` API. I runt it without problems, so I think we are ok here.

PR #116, opened 2024-10-30
Port ZipStreamReader from adamziel/wxr-normalize

adamziel:
This new ZipStreamReader opens its own file handles which means it can be paused, resumed, and is more reliable. The original implementation was built as a part of https://github.com/adamziel/wxr-normalize/pull/1

This is all new code so there are no testing instructions. Eventually this implementation will replace the existing ZipStreamReader.


adamziel:
The failing tests seem to be also an issue with phpunit setup. I'll go ahead and merge anyway.

PR #133, opened 2025-07-24
Make WordPress polyfills optional

adamziel:
`php-toolkit` relies on WordPress APIs, such as `apply_filters()`. Sometimes they're missing, e.g. in PHPUnit runtime, and they're implicitly polyfilled when including `vendor/autoload.php`:

```php
if ( ! class_exists( 'WP_Exception' ) ) {
	class WP_Exception extends Exception {
	}
}
```

This polyfilling technique assumes the library is either:

* Loaded inside a WordPress plugin, where WP_Exception is already available
* Loaded outside of WordPress, where there's no WP_Exception class

However, it does not account an important third scenario that [WooCommerce ran into](https://github.com/woocommerce/woocommerce/actions/runs/16489847925/job/46621900586?pr=59851):

* Loaded in a WordPress runtime, but before `wp-load.php` is included

## Solution

With this PR, polyfills are no longer implicitly loaded as a part of the autoload sequence. Instead, we expose an explicit `polyfill_wordpress_apis()` function that must be called by the library consumer.

### Rationale

At the time of `require "vendor/autoload.php"`, we don't know what the developer is going to do next. Maybe they'll include `wp-load.php`, in which case we must not load the polyfills. Or maybe then won't include `wp-load.php`, in which case we must load the polyfills. Exposing an explicit `polyfill_wordpress_apis()` function lets them decide.

The one downside is: it creates a gotcha. Anyone who wants to consume `php-toolkit` as a library must know about that function and must call it, otherwise they'll see a fatal error. I'm noodling on adding an explicit warning to communicate this problem directly in the CLI without an additional google search.

## Testing plan

Confirm the CI tests pass – many `php-toolkit` PHPUnit tests rely on WordPress polyfills and won't work if `polyfill_wordpress_apis()` is not called or is incomplete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,15 @@ You can also lock it in to a specific commit or tag:
 }
 ```
 
+If you are running the libraries in a non-WordPress environment, e.g. a PHPUnit test suite that
+tests your code without loading WordPress, you'll need to polyfill the WordPress APIs used by
+the php-toolkit libraries:
+
+```php
+require_once __DIR__ . '/vendor/autoload.php';
+polyfill_wordpress_apis();
+```
+
 For now, there is no way to cherry-pick just the one library you need. It's all or nothing.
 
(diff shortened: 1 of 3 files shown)


naman03malhotra:
@adamziel commenting here to be assigned as a reviewer.

PR #136, opened 2025-08-01
[Blueprints v2] Add support for "enableMultisite" step

### LINKED ISSUE #130: [Blueprints v2] Support enableMultisite step
Blueprints v1 support multisite creation. Blueprints v2 do not. The multisite step is actually somewhat popular – let's make sure it's supported for the initial Blueprints v2 rollout in Playground.

cc @janjakes

JanJakes:
Resolves https://github.com/WordPress/php-toolkit/issues/130.
Closes https://github.com/WordPress/php-toolkit/pull/131.

### DIFF
--- a/components/Blueprints/Steps/EnableMultisiteStep.php
+++ b/components/Blueprints/Steps/EnableMultisiteStep.php
@@ -1,0 +1,114 @@
+<?php
+
+namespace WordPress\Blueprints\Steps;
+
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use WordPress\Blueprints\Exception\BlueprintExecutionException;
+use WordPress\Blueprints\Progress\Tracker;
+use WordPress\Blueprints\Runtime;
+
+/**
+ * Represents the 'enableMultisite' step.
+ */
+class EnableMultisiteStep implements StepInterface {
+	public function run( Runtime $runtime, Tracker $tracker ) {
+		$tracker->setCaption( 'Enabling multisite' );
+
+		$code =
+<<<'PHP'
+<?php
+/*
+ * This code is mirroring the "wp core multisite-convert" command behavior.
+ * See: https://github.com/wp-cli/core-command/blob/f157fb37dae1d13fe7318452f932917161e83e53/src/Core_Command.php#L505
+ */
+
+require_once getenv( 'DOCROOT' ) . '/wp-load.php';
+require_once getenv( 'DOCROOT' ) . '/wp-admin/includes/upgrade.php';
+
+// need to register the multisite tables manually for some reason
+foreach ( $wpdb->tables( 'ms_global' ) as $table => $prefixed_table ) {
+	$wpdb->$table = $prefixed_table;
+}
+
+install_network();
+
+// Get multisite arguments
+$site_id     = 1;
+$base        = '/';
+$title       = sprintf( '%s Sites', get_option( 'blogname' ) );
+$admin_email = get_option( 'admin_email' );
+$subdomains  = false;
+
+// Get the base domain
+$siteurl = get_option( 'siteurl' );
+$domain  = (string) preg_replace( '|https?://|', '', $siteurl );
+$slash   = strpos( $domain, '/' );
+if ( false !== $slash ) {
+	$domain = substr( $domain, 0, $slash );
+}
+
+// Eagerly check for custom ports
+if ( strpos( $domain, ':' ) !== false ) {
+	throw new Exception(
+		sprintf(
+			'The current host is "%s", but WordPress multisites do not support custom ports.',
+			$domain
+		)
+	);
+}
+
+$result = populate_network(
+	$site_id,
+	$domain,
+	$admin_email,
+	$title,
+	$base,
+	$subdomains
+);
+
+$site_id = $wpdb->get_var( "SELECT id FROM $wpdb->site" );
+$site_id = ( null === $site_id ) ? 1 : (int) $site_id;
+
+if ( $result instanceof WP_Error ) {
+	throw new Exception(
+		sprintf(
+			'Error: [%s] %s',
+			$result->get_error_code(),
+			$result->get_error_message()
+		)
+	);
+}
+
+// delete_site_option() cleans the alloptions cache to prevent dupe option
+delete_site_option( 'upload_space_check_disabled' );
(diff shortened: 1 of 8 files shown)


components/Blueprints/Steps/EnableMultisiteStep.php
```diff
@@ -0,0 +1,91 @@
...
+
+namespace WordPress\Blueprints\Steps;
+
+use WordPress\Blueprints\Progress\Tracker;
+use WordPress\Blueprints\Runtime;
+
+/**
+ * Represents the 'enableMultisite' step.
+ */
+class EnableMultisiteStep implements StepInterface {
```
adamziel, line 13: Oh, nice! You got it to work without actually depending on wp-cli!

adamziel:
It's a promising start @JanJakes, thank you! We could use a couple of unit tests here – LMK if you'd like to connect on how to build them.

JanJakes:
@adamziel I'm just finishing the tests; I will push it soon. Should the behavior be as close to the v1 behavior as possible? For example, when a site is already multisite, should it fail with `This already is a multisite installation.`, as v1 does?

adamziel:
> Should the behavior be as close to the v1 behavior as possible? For example, when a site is already multisite, should it fail with `This already is a multisite installation`, as v1 does?

For now, yes. It may be cumbersome in some cases but could prevent catastrophic failures in other cases.

JanJakes:
There are some interesting test failures in the macOS and Win builds, while Ubuntu passes. It's likely related to how some `$_SERVER` variables are set by WP CLI. I'll check it out further.

JanJakes:
@adamziel This should be ready now. The test failures seem to be unrelated/flaky.

adamziel:
Nitpick: Spaces vs tabs – the comments are visually less indented than the lines they refer to

CleanShot 2025-08-05 at 18 03 09@2x

adamziel (approved):
Really good work @JanJakes! My only notes are about code formatting – feel free to merge once that's sorted out!

JanJakes:
@adamziel There's one more thing that came to my mind: The `populate_network()` function calls `wp_guess_url()` and that relies on some `$_SERVER` variables like `$_SERVER['HTTP_HOST']` and `$_SERVER['REQUEST_URI']` being set. I think this affects which site URL the main site in the network will have set.

When using the CLI via `php bin/blueprint.php exec`, these may be required to be set based on the `--site-url` parameter. It seems that [WP CLI does something like that](https://github.com/wp-cli/wp-cli/blob/a499fe6c93eab0ee48b660841fe81ec5f311ab31/php/class-wp-cli.php#L129), so we may mirror the behavior.

I will do some testing via the `php bin/blueprint.php exec` command and compare the behavior with WP CLI.

JanJakes:
@adamziel OK, so particularly for the enable-multisite path, the `$_SERVER` values are not important as the `populate_network` path seems to be fetching the `siteurl` option for both [the network](https://github.com/WordPress/php-toolkit/pull/136/commits/1de041a819415ad47c8f05f9a280eec320b290ae#diff-6bf8489ee05798405a18f0ebdeac0282298bdb8ac6bd9198644b4fcd10625041R43) and [the initial blog](https://github.com/WordPress/wordpress-develop/blob/c726220a21d13fdb5409372b652c9460c59ce1db/src/wp-admin/includes/schema.php#L1275C1-L1276C1).

That said, would it be safer to add something like the below to the subprocess script prefix? It's [as per WP CLI](https://github.com/wp-cli/wp-cli/blob/a499fe6c93eab0ee48b660841fe81ec5f311ab31/php/class-wp-cli.php#L129).
Screenshot 2025-08-06 at 11 51 05

adamziel:
@JanJakes would we be adding that just in case? Or did you run into an actual case where lacking those values is an issue? I'm worried about confusing WordPress into thinking we're running via a CGI SAPI when we're actually running as CLI.

JanJakes:
> @JanJakes would we be adding that just in case? Or did you run into an actual case where lacking those values is an issue? I'm worried about confusing WordPress into thinking we're running via a CGI SAPI when we're actually running as CLI.

@adam At this stage, it only seems to be a precaution. It doesn't affect the multisite step as it is, and likely not other steps we have now. What's a bit tricky, though, is that in many of the steps we call `wp-load.php`, and some of the WP functionality may rely on these constants. [WP CLI sets these as per the `--url` parameter](https://github.com/wp-cli/wp-cli/blob/a499fe6c93eab0ee48b660841fe81ec5f311ab31/php/WP_CLI/Runner.php#L1233) and we have a mandatory `--site-url` parameter that's actually passed to WP CLI for the WP installation but not set in this way in our steps.

adamziel (approved):

JanJakes:
> @adam At this stage, it only seems to be a precaution. It doesn't affect the multisite step as it is, and likely not other steps we have now. What's a bit tricky, though, is that in many of the steps we call wp-load.php, and some of the WP functionality may rely on these constants. [WP CLI sets these as per the --url parameter](https://github.com/wp-cli/wp-cli/blob/a499fe6c93eab0ee48b660841fe81ec5f311ab31/php/WP_CLI/Runner.php#L1233) and we have a mandatory --site-url parameter that's actually passed to WP CLI for the WP installation but not set in this way in our steps.

@adamziel Just for the record, I didn't add this. If you'd prefer to do it, I can create a new PR.

adamziel:
I'm on the fence. I looked up why WP-CLI does that but the decisions are not well documented. It seems like that's just their way of supporting the `--blog` and `--url` CLI options, not something inherently required by WP:

* https://github.com/wp-cli/wp-cli/commit/ed8f88770f581683f5c45d98594fec87250fd7fc
* https://github.com/wp-cli/wp-cli/issues/69
* https://github.com/wp-cli/wp-cli/commit/8d96313fe2c031eb0ac228de6d391669d72bdbc0

Lacking these attributes didn't give me any troubles yet. At the same time, when I had them set, I occasionally saw WordPress trying to send the `Location: ` header from a CLI script.

JanJakes:
@adamziel We can also wait and see if it's needed at some point. In WP CLI, I noticed it seems to be needed for multisite install but not multisite conversion (as we're doing). There are some cases when the CLI commands land on `wp_guess_url()` which needs these variables. In general, I think WP Core is not really built with CLI in mind, so you can definitely run into `header()` calls that will be a noop on the CLI.

PR #137, opened 2025-08-04
[WIP] Streaming WXR importer plugin

adamziel:
This is an exploration of an alternative, streaming-first WordPress importer plugin. It ships an Interactivity API-based UI and supports importing large WXR files without timing out or running out of memory.

This PR ships a few vital improvements for the php-toolkit library around logging and tracking progress, but is not intended for merging as is. Instead, I've meant it as a "5 steps ahead" explorations to reveal any roadblocks we may run into when [improving the official WXR importer plugin](https://github.com/WordPress/php-toolkit/issues/138).


zaerl (changes requested):
I left some comments, tell me what you think.

PR #142, opened 2025-08-08
[Blueprints v2] Constraint the Blueprint bundle format

### LINKED ISSUE #132: [Blueprints v2] Constraint the Blueprint bundle format
The Blueprints v2 spec defines a [required bundle directory structure](https://github.com/Automattic/WordPress-extension-proposals/blob/trunk/wep-1-blueprint-v2-schema/proposal.md#bundle-directory-structure). Let's make sure any bundle not following this structure is rejected by the PHP runner. cc @janjakes

```
blueprint.json
wp-content/
└─── plugins/
|    └─── akismet.zip
|    └─── custom-meta-box.php
|    └─── hello-dolly/
|         └─── hello-dolly.php
└─── mu-plugins/
|    └─── 0-check-integrity.zip
└─── themes/
|    └─── twentytwentythree/
|         └─── style.css
|         └─── functions.php
|         └─── screenshot.jpg
└─── languages/
|    └─── plugins/
|         └─── akismet/
|              └─── akismet-en_US.po
|              └─── akismet-en_US.mo
|              └─── akismet-de_DE.po
|              └─── akismet-de_DE.mo
|    └─── themes/
|         └─── twentytwentythree/
|              └─── twentytwentythree-en_US.po
|              └─── twentytwentythree-en_US.mo
|    └─── en_US.po
|    └─── en_US.mo
|    └─── de_DE.po
|    └─── de_DE.mo
└─── uploads/
|    └─── fonts/
|         └─── open-sans.woff2
|         └─── roboto.woff2
|         └─── ...
|    └─── 2024/
|         └─── 01/
|              └─── image.jpg
|              └─── video.mp4
|              └─── audio.mp3
|              └─── document.pdf
|              └─── other.txt
└─── content/
     └─── dump-1.sql
     └─── dump-2.sql
     └─── dump-3.sql
     └─── bike-repair-shop.wxr
     └─── posts/
           └─── posts/
				└─── 2025-01-01-my-first-post.html
           └─── pages/
                └─── books/
                     └─── post-type.json
                     └─── the-old-man-and-the-sea.html
                     └─── harry-potter.md
```

JanJakes:
WIP

Closes https://github.com/WordPress/php-toolkit/issues/132.

### DIFF
--- a/components/Blueprints/Runner.php
+++ b/components/Blueprints/Runner.php
@@ -43,6 +43,7 @@ use WordPress\Blueprints\Steps\UnzipStep;
 use WordPress\Blueprints\Steps\WPCLIStep;
 use WordPress\Blueprints\Steps\WriteFilesStep;
 use WordPress\Blueprints\Validator\HumanFriendlySchemaValidator;
+use WordPress\Blueprints\Validator\ValidationError;
 use WordPress\Blueprints\Versions\Version1\V1ToV2Transpiler;
 use WordPress\Blueprints\VersionStrings\PHPVersion;
 use WordPress\Blueprints\VersionStrings\VersionConstraint;
@@ -166,7 +167,7 @@ class Runner {
 		}
 
 		// Validate database credentials
-		$dbCreds = $config->getDatabaseCredentials();
+		/*$dbCreds = $config->getDatabaseCredentials();
 		if ( $dbEngine === 'mysql' ) {
 			if ( empty( $dbCreds['username'] ) || empty( $dbCreds['databaseName'] ) ) {
 				throw new BlueprintExecutionException( "MySQL credentials are required when database engine is 'mysql'." );
@@ -197,7 +198,7 @@ class Runner {
 			if ( empty( $dbCreds['path'] ) ) {
 				$dbCreds['path'] = 'wp-content/.ht.sqlite';
 			}
-		}
+		}*/
 	}
 
 	public function run(): void {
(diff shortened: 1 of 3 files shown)


components/Blueprints/Runner.php
```diff
@@ -402,127 +403,22 @@ private function validateBlueprint(): void {
...
-			// NewSiteResolver and ExistingSiteResolver.
-		}
-
-		// Validate the override constraint if it was set
-		if ( $this->wpVersionConstraint ) {
-			$wpConstraintErrors = $this->wpVersionConstraint->validate();
-			if ( ! empty( $wpConstraintErrors ) ) {
-				throw new BlueprintExecutionException( 'Invalid WordPress version constraint from CLI override: ' . implode( '; ', $wpConstraintErrors ) );
-			}
+		$validator = new \WordPress\Blueprints\Validator\BlueprintValidator( $this->blueprintArray, $this->blueprintExecutionContext );
```
adamziel, line 406: Nice idea separating it! I've been experimenting with co-locating that logic with parsing the Blueprint and it had some ups and downs.

The downside here is that it's separated from parsing the Blueprint. The upside here is that it's separated from parsing the Blueprint. Let's roll with a separate validator and see how it goes.

components/Blueprints/Validator/BlueprintValidator.php
```diff
@@ -0,0 +1,573 @@
...
+
+    private function validateWpContent(Filesystem $fs): ?ValidationError {
+        if (!$fs->exists('/wp-content')) {
+            return null;
+        }
+
+        $errors = [];
+        $rootPointer = '/bundle/wp-content';
+
+        // Top-level allowed directories within wp-content
```
adamziel, line 218: I wonder if there's a reason to allow arbitrary directories 🤔 As in, themes, langs etc. would still have to be in specific paths, but would it be useful to allow, say, `my-custom-dir` here? I don't know, it's an open question – I wonder if there are sites or plugins out there that rely on those. If yes, we can adjust the spec.

Also – it's not blocking here. If anyone reports a compelling use-case later on, we can easily relax this constraint.

components/Blueprints/Validator/BlueprintValidator.php
```diff
@@ -0,0 +1,573 @@
...
+        if ($fs->exists('/wp-content/languages') && $fs->is_dir('/wp-content/languages')) {
+            foreach ($fs->ls('/wp-content/languages') as $entry) {
+                $entryPath = '/wp-content/languages/' . $entry;
+                $pointer = $rootPointer . '/languages/' . $entry;
+                if ($fs->is_dir($entryPath)) {
+                    if (!in_array($entry, ['plugins', 'themes'], true)) {
+                        $errors[] = new ValidationError(
+                            $pointer,
+                            'unexpected-language-subdir',
+                            sprintf('Unexpected directory under languages: %s. Allowed: plugins/, themes/, or .po/.mo files at root.', $entry)
```
adamziel, line 285: Would .pot/.mot be useful here? Or not? Any .json files we expect? Again, I only have questions, not answers.

adamziel: Perhaps we should tolerate any files here, even .png, to support cases where you just zip a WordPress site and call that a Blueprint. However, if you want to refer to a translation file from Blueprint.json, that file must live in wp-content/languages. If I'm contradicting the Blueprints spec here, let's discuss and adjust the spec if needed.

adamziel: So what I'm saying is that the validation is more about confirming the .json file refers to allowed paths than about constraining what is allowed in those paths. In a scenario where we treat a URL as a Blueprint, say https://adamziel.com/my-demo-site/, we don't even have the ability to list files inside a directory.

components/Blueprints/Runner.php
JanJakes: > The downside here is that it's separated from parsing the Blueprint. The upside here is that it's separated from parsing the Blueprint.

This. I think that if we can make the parsing logic collect all errors before bailing out while keeping a reasonably nice code, then we should go with that. I'm looking into that.

components/Blueprints/Validator/BlueprintValidator.php
JanJakes: Thanks for jumping in! Indeed, this validation is very strictly focused on the directory structure, without looking at the references in the JSON file. I think it will be much better to only validate if the referenced paths are correct and then only output some warning that there are some unused (non-referenced) files.

JanJakes:
I'm exploring an alternative with more Blueprint parsing & validation collocation in https://github.com/WordPress/php-toolkit/pull/144.

PR #145, opened 2025-08-15
Fix flaky tests and errors discovered by them

JanJakes:
#### 1. Fix `WordPress\ByteStream\ByteStreamException: Cannot seek to a negative offset`
This error seemed to occur in some scenarios when the `RequestReadStream` finished reading data without correctly reaching the end of data, leaving a `FINISHED` event unprocessed. This caused no data length to be set for requests without the `Content-Length` header, because this logic is in the `FINISHED` even handler.

I fixed it to make sure the remote stream is always consumed until the end of data.

Locally on macOS, it happens rarely. To reproduce it, I used the command below:
```bash
while php ./vendor/bin/phpunit -c phpunit.xml --filter testReadRemoteZip components/Zip/Tests/ZipFilesystemTest.php; do :; done
```
After up to a few dozen iterations, the error appears.

#### 2. Fix `stream_socket_enable_crypto(): SSL operation failed with code 1.`
The full error message is:
```bash
stream_socket_enable_crypto(): SSL operation failed with code 1. OpenSSL Error messages:
error:0A0000C6:SSL routines::packet length too long
error:0A000139:SSL routines::record layer failure
```

This is a warning [emitted by `stream_socket_enable_crypto()`](https://github.com/WordPress/php-toolkit/blob/9b032ae571e68264efc0ce09a144185d755d0869/components/HttpClient/Transport/SocketTransport.php#L252) that is then converted by PHPUnit to a `PHPUnit\Framework\Error\Warning` error.

Again, it only happens sometimes, and it is a mystery to me why. Here's a way to reproduce it on `macOS`:
```bash
while php ./vendor/bin/phpunit -c phpunit.xml --filter test_ssl_handshake_failure components/HttpClient/Tests/SocketTransportTest.php; do :; done
```
After up to a few dozen iterations, the error appears.

#### 3. Fix `HttpMiddleware` timeout sometimes cutting off `SocketTransport` timeout
It seems that `WordPress\HttpClient\Transport\SocketTransport` sometimes times out a bit later than the implementation of `WordPress\HttpClient\Middleware\HttpMiddleware` was expecting, which was causing some flaky tests. Adding a minimum of 300ms extra time in `HttpMiddleware` seems to solve the issue.

Here's a way to reproduce it on `macOS`:
```bash
while php ./vendor/bin/phpunit -c phpunit.xml --filter test_errors components/HttpClient/Tests/SocketTransportTest.php; do :; done
```
After up to a few dozen iterations, the error appears. It also helps to comment out [all other data provider cases](https://github.com/WordPress/php-toolkit/blob/9b032ae571e68264efc0ce09a144185d755d0869/components/HttpClient/Tests/SocketTransportTest.php#L148-L154) and decrease the [client timeout](https://github.com/WordPress/php-toolkit/blob/9b032ae571e68264efc0ce09a144185d755d0869/components/HttpClient/Tests/SocketTransportTest.php#L129), which makes it more flaky (10% extra time of a smaller number means a smaller margin).

#### 4. `SocketTransportTest::test_redirect_chain`
The error message in the tests was the following:
```
1) WordPress\HttpClient\Tests\SocketTransportTest::test_redirect_chain
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Redirect 2'
+''
```

It appears that sometimes, when a request was transitioning through the `STATE_RECEIVED` state, the buffered body could be lost in `ClientState:: consume_buffered_response_body()`. Adding the state to a condition in that method seems to have resolved the issue. I wonder if any more states need to be added there, or if now the list is complete.

#### 5. Fix cache invalidation on Windows with PHP <= 7.2

The `WordPress\HttpClient\Tests\CacheMiddlewareIntegrationTest` was failing on Windows with PHP 7.2 with the following errors:

```
1) WordPress\HttpClient\Tests\CacheMiddlewareIntegrationTest::test_post_invalidates_cache
Failed asserting that an array is empty.

D:\a\php-toolkit\php-toolkit\components\HttpClient\Tests\CacheMiddlewareIntegrationTest.php:271
D:\a\php-toolkit\php-toolkit\components\HttpClient\Tests\WithServerTrait.php:31
D:\a\php-toolkit\php-toolkit\components\HttpClient\Tests\CacheMiddlewareIntegrationTest.php:272
phpvfscomposer://D:\a\php-toolkit\php-toolkit\vendor\phpunit\phpunit\phpunit:97

2) WordPress\HttpClient\Tests\CacheMiddlewareTest::test_non_cacheable_methods_invalidate_cache
Failed asserting that an array is empty.

D:\a\php-toolkit\php-toolkit\components\HttpClient\Tests\CacheMiddlewareTest.php:69
phpvfscomposer://D:\a\php-toolkit\php-toolkit\vendor\phpunit\phpunit\phpunit:97
```

As it turns out, before PHP 7.3 it was not possible to `unlink()` a file with handles in use. I added a condition to address that.

#### 6. Tests failing when calling `$this->internal_close_reading();` in `BaseByteReadStream`
When debugging the test failures, I noticed the following snippet in `BaseByteReadStream`:
```php
public function close_reading(): void {
	$this->is_read_closed = true;
	// @TODO: Closing the underlying stream here breaks the tests. Why?
	// $this->internal_close_reading();
	$this->internal_close_reading();
}
```

I uncommented the line and encountered two failures:

1. In `ZipFilesystemTest::testReadFile` I got `TypeError: fseek(): Argument #1 ($stream) must be of type resource, null given`. This happened because `LimitedByteReadStream.php` was calling `$this->upstream->close_reading();`. Looking at other similar stream classes, it seemed to me that this should in fact not be called, as the class doesn't initialize the upstream. In other cases, we also don't call `close_reading()` on the upstream. I hope that's the correct fix.
2. In `SeekableRequestReadStreamTest::testCloseReading`, I was sometimes getting `ByteStreamException: Cancelling the request is not implemented yet`. I think this is expected, so I made the test expect the error and made the exception to always appear.

### DIFF
--- a/components/ByteStream/ReadStream/LimitedByteReadStream.php
+++ b/components/ByteStream/ReadStream/LimitedByteReadStream.php
@@ -40,8 +40,4 @@ class LimitedByteReadStream extends BaseByteReadStream {
 	protected function seek_outside_of_buffer( int $target_offset ): void {
 		$this->upstream->seek( $this->initial_offset + $target_offset );
 	}
-
-	protected function internal_close_reading(): void {
-		$this->upstream->close_reading();
-	}
 }
--- a/components/HttpClient/ByteStream/SeekableRequestReadStream.php
+++ b/components/HttpClient/ByteStream/SeekableRequestReadStream.php
@@ -105,10 +105,9 @@ class SeekableRequestReadStream implements ByteReadStream {
 	public function consume_all(): string {
 		while ( ! $this->remote->reached_end_of_data() ) {
 			$pulled = $this->remote->pull( BaseByteReadStream::CHUNK_SIZE );
-			if ( 0 === $pulled ) {
-				break;
+			if ( $pulled > 0 ) {
+				$this->cache->append_bytes( $this->remote->consume( $pulled ) );
 			}
-			$this->cache->append_bytes( $this->remote->consume( $pulled ) );
 		}
 		$this->cache->close_writing();
 
--- a/components/HttpClient/ClientState.php
+++ b/components/HttpClient/ClientState.php
@@ -178,6 +178,7 @@ class ClientState {
 		$connection = $this->connections[ $request->id ];
 		if (
 			$request->state === Request::STATE_RECEIVING_BODY ||
+			$request->state === Request::STATE_RECEIVED ||
 			$request->state === Request::STATE_FINISHED
 		) {
 			return $connection->consume_buffer();
--- a/components/HttpClient/Middleware/CacheMiddleware.php
+++ b/components/HttpClient/Middleware/CacheMiddleware.php
@@ -61,13 +61,15 @@ final class CacheMiddleware implements MiddlewareInterface {
 
 	public function await_next_event( $requests_ids ): bool {
 		/* serve cached replay first */
-		foreach ( $this->replay as $id => $context ) {
-			if ( $context['done'] ) {
-				fclose( $context['file'] );
-				unset( $this->replay[ $id ] );
-				continue;
-			}
+		$id = array_keys( $this->replay )[0] ?? null;
+		if ( null !== $id ) {
 			$this->fromCache( $id );
+
+			// When done, close the file handle and clean up the replay entry.
+			if ( $this->replay[ $id ]['done'] ) {
+				fclose( $this->replay[ $id ]['file'] );
+				unset( $this->replay[ $id ] );
+			}
 			return true;
 		}
 		
@@ -413,9 +415,18 @@ final class CacheMiddleware implements MiddlewareInterface {
 		if ( $fp = @fopen( $metaFile, 'c' ) ) {
 			flock( $fp, LOCK_EX );
 		}
+
 		// Delete cache files if they exist
 		@unlink( $bodyFile );
+		if ( 'Windows' === PHP_OS_FAMILY && PHP_VERSION_ID < 70300 && isset( $fp ) && $fp ) {
+			// On Windows, PHP 7.2 or earlier can't "unlink()" files with handles in use.
+			// In that case, we need to first release the lock, and then unlink the file.
+			flock( $fp, LOCK_UN );
+			fclose( $fp );
+			unset( $fp );
+		}
 		@unlink( $metaFile );
+
 		// Also remove any temp files for this entry
 		foreach ( glob( $bodyFile . '.tmp*' ) as $tmp ) {
 			@unlink( $tmp );
--- a/components/HttpClient/Middleware/HttpMiddleware.php
+++ b/components/HttpClient/Middleware/HttpMiddleware.php
@@ -112,11 +112,9 @@ class HttpMiddleware implements MiddlewareInterface {
 		$this->state->request             = null;
 		$this->state->response_body_chunk = null;
 
+		// Give the requests an opportunity to time out; 10% more, but at least 300ms.
+		$timeout_ms = $this->state->request_timeout_ms + max( 300, $this->state->request_timeout_ms * 0.1 );
 		$start_time = microtime( true );
-		$timeout_ms = isset( $query['timeout_ms'] )
-			? $query['timeout_ms']
-			// Give the requests an opportunity to time out
-			: $this->state->request_timeout_ms * 1.1;
 
 		do {
 			foreach ( $requests_ids as $request_id ) {
--- a/components/HttpClient/Tests/SocketTransportTest.php
+++ b/components/HttpClient/Tests/SocketTransportTest.php
@@ -131,7 +131,7 @@ class SocketTransportTest extends ClientTestBase {
             $client->enqueue( $request );
 
             $error_occurred = false;
-            while ( $client->await_next_event( [ 'requests' => [ $request ], 'timeout_ms' => 2000 ] ) ) {
+            while ( $client->await_next_event( [ 'requests' => [ $request ] ] ) ) {
                 switch ( $client->get_event() ) {
                     case Client::EVENT_FAILED:
                         $error_occurred = true;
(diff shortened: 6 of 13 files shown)


components/ByteStream/ReadStream/LimitedByteReadStream.php
```diff
@@ -40,8 +40,4 @@ protected function internal_reached_end_of_data(): bool {
 	protected function seek_outside_of_buffer( int $target_offset ): void {
 		$this->upstream->seek( $this->initial_offset + $target_offset );
 	}
-
-	protected function internal_close_reading(): void {
-		$this->upstream->close_reading();
-	}
```
JanJakes, line 46: I hope this fix is correct. Looking at other similar stream classes, we don't seem to `close_reading()` on the upstream unless the class initialized it.

components/HttpClient/ByteStream/SeekableRequestReadStream.php
```diff
@@ -105,10 +105,9 @@ public function consume( $n ): string {
 	public function consume_all(): string {
 		while ( ! $this->remote->reached_end_of_data() ) {
 			$pulled = $this->remote->pull( BaseByteReadStream::CHUNK_SIZE );
-			if ( 0 === $pulled ) {
-				break;
+			if ( $pulled > 0 ) {
+				$this->cache->append_bytes( $this->remote->consume( $pulled ) );
 			}
-			$this->cache->append_bytes( $this->remote->consume( $pulled ) );
```
JanJakes, line 111: This is just to be sure we always run until `reached_end_of_data()`. It wasn't the actual problem, but the code looked like it could've been the culprit at first. [The actual fix is here](https://github.com/WordPress/php-toolkit/pull/145/files#diff-ca4392776b55c0e21d14edd54c25fa6ce88091c2fa5e02c4be3063e361d50c82R208).

components/HttpClient/Middleware/CacheMiddleware.php
```diff
@@ -61,13 +61,15 @@ public function enqueue( Request $request ) {
...
+		if ( null !== $id ) {
+			$this->fromCache( $id );
+
+			// When done, close the file handle and clean up the replay entry.
+			if ( $this->replay[ $id ]['done'] ) {
+				fclose( $this->replay[ $id ]['file'] );
 				unset( $this->replay[ $id ] );
-				continue;
 			}
-			$this->fromCache( $id );
```
JanJakes, line 70: Here I only switch the order of `fromCache()` and `done` state cleanup. Before, it would be cleaned up on the next `await_next_event`; having it cleaned right after the `done` event seemed better, and it eliminated the need for the loop and `continue` logic. [The actual fix is here](https://github.com/WordPress/php-toolkit/pull/145/files#diff-9f7d89840b11798fb19e25854695da98b46e1d872d1af08972a568cb87e56972R421).

I also noticed the `$requests_ids` parameter is not used in the caching part, but I didn't dig further to figure out whether that's fine. I suppose it is?

components/HttpClient/Middleware/HttpMiddleware.php
```diff
@@ -112,11 +112,9 @@ public function await_next_event( $requests_ids ) :bool {
 		$this->state->request             = null;
 		$this->state->response_body_chunk = null;
 
+		// Give the requests an opportunity to time out; 10% more, but at least 300ms.
+		$timeout_ms = $this->state->request_timeout_ms + max( 300, $this->state->request_timeout_ms * 0.1 );
 		$start_time = microtime( true );
-		$timeout_ms = isset( $query['timeout_ms'] )
-			? $query['timeout_ms']
-			// Give the requests an opportunity to time out
-			: $this->state->request_timeout_ms * 1.1;
```
JanJakes, line 119: The `$query['timeout_ms']` was unused and not passed in.

components/HttpClient/Tests/SocketTransportTest.php
```diff
@@ -131,7 +131,7 @@ public function test_errors( $scenario, $expectedErrorSubstring ) {
             $client->enqueue( $request );
 
             $error_occurred = false;
-            while ( $client->await_next_event( [ 'requests' => [ $request ], 'timeout_ms' => 2000 ] ) ) {
+            while ( $client->await_next_event( [ 'requests' => [ $request ] ] ) ) {
```
JanJakes, line 134: The `timeout_ms` is not used in the target class.

components/HttpClient/ClientState.php
```diff
@@ -178,6 +178,7 @@ public function consume_buffered_response_body( $request_id ) {
 		$connection = $this->connections[ $request->id ];
 		if (
 			$request->state === Request::STATE_RECEIVING_BODY ||
+			$request->state === Request::STATE_RECEIVED ||
```
JanJakes, line 181: Are there any other states that should be considered here?

akirk:
Can we also fix the error that comes with a default `composer install`:
```
PHP Fatal error:  Cannot redeclare function WordPress\DataLiberation\URL\wp_rewrite_urls() (previously declared in php-toolkit/components/DataLiberation/URL/functions.php:31) in php-toolkit/components/DataLiberation/URL/functions.php on line 31
```

I believe the problem is the autoload in composer.json. If you remove it afterwards you need to run `composer dump-autoload` to make it work.

JanJakes:
> Can we also fix the error that comes with a default composer install:

@akirk It turned out to be just a `composer.lock` that was out-of-sync with the latest `composer.json`. Previously, `wordpress/data-liberation` was symlinked in `vendor`, now it's loaded from the root using `PSR-4` and `files`. Running `composer update` fixed the issue.

components/ByteStream/ReadStream/LimitedByteReadStream.php
adamziel: Makes sense, I think it's a leftover from the previous shape of this API that involved chaining those calls

components/HttpClient/Middleware/CacheMiddleware.php
adamziel: > I also noticed the $requests_ids parameter is not used in the caching part, but I didn't dig further to figure out whether that's fine. I suppose it is?

It should be fine, but good spot!

components/HttpClient/Middleware/CacheMiddleware.php
```diff
@@ -413,9 +415,18 @@ public function invalidateCache( Request $request ): void {
 		if ( $fp = @fopen( $metaFile, 'c' ) ) {
 			flock( $fp, LOCK_EX );
 		}
+
 		// Delete cache files if they exist
 		@unlink( $bodyFile );
+		if ( 'Windows' === PHP_OS_FAMILY && PHP_VERSION_ID < 70300 && isset( $fp ) && $fp ) {
+			// On Windows, PHP 7.2 or earlier can't "unlink()" files with handles in use.
```
adamziel, line 422: Oooh great find

components/HttpClient/ClientState.php
adamziel: Can't think of any

adamziel:
Great work here @JanJakes, you've got so many things right in a lot of low-level areas – mad kudos.

PR #149, opened 2025-08-30
XMLProcessor and ByteStream improvements for wordpress-importer

adamziel:
Address @dmsnell's feedback from https://github.com/WordPress/wordpress-importer/pull/190:

> this looks nice. I believe that the curly brackets are illegal in tag names, right? so this is a way we can encoding the namespace of tag? I think you explained this to me before but I wanted to leave a note here about it.

([link](https://github.com/WordPress/wordpress-importer/pull/190#discussion_r2298983774))

curly brackets are indeed illegal in tag names. I've added an "Internal representation of names" section to the comment block to clarify this and refer the relevant XML spec fragment.

> units would be helpful, and the comment seems to directly contradict the code.

([link](https://github.com/WordPress/wordpress-importer/pull/190#discussion_r2298986766))

I've increased the CHUNK_SIZE to 64kb and renamed it to CHUNK_SIZE_BYTES

> the only way to embed CDATA in an element is inside a CDATA section with <![CDATA[…]]>.

([link](https://github.com/WordPress/wordpress-importer/pull/190#discussion_r2299261462))

I've removed the entire `declare_element_as_pcdata` semantics.

> $buffer_size is given as 2048 but this appears to be pulling self::CHUNK_SIZE, which does not match. should they match?

([link](https://github.com/WordPress/wordpress-importer/pull/190#discussion_r2298990175))

I've renamed that variable to `$max_lookbehind_bytes` and documented it to clarify the meaning

## Testing

CI tests should continue to pass.

cc @dmsnell

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,3 @@
-# Repository Archived
-
-The development continues in https://github.com/WordPress/php-toolkit
-
----
-
 ## PHP Toolkit
 
 Standalone PHP libraries for use in WordPress plugins and standalone PHP projects:
--- a/README.md
+++ b/README.md
@@ -1,3 +1,9 @@
+# Repository Archived
+
+The development continues in https://github.com/WordPress/php-toolkit
+
+--
+
 ## PHP Toolkit
 
 Standalone PHP libraries for use in WordPress plugins and standalone PHP projects:
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 The development continues in https://github.com/WordPress/php-toolkit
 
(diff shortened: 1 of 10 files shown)


adamziel:
This is mostly comments, variable renaming, and removing code. All the tests pass, I'll go ahead and merge. CCing @JanJakes

dmsnell:
nice. thanks.

`lookbehind` connotes something _more_ to me, but I think it sounds good. it’s much clearer than `buffer`.

PR #154, opened 2025-09-05
Apply PHPCBF to all components

adamziel:
Applies PHPCBF to the source code in this repo. There's a bunch of "violations" left – TBD whether we'll disable those rules or adjust the code to fulfill them.


adamziel:
It seems like some sniffs are introducing syntax errors, e.g. this:

```php
[ $prop, $allowed ] = $disc;
```

Was transformed to this:

```php
array( $prop, $allowed ) = $disc;
```

PHPCBF really likes the `array()` syntax 🤦

adamziel:
Let's take it in smaller bites, starting with #158

PR #155, opened 2025-09-05
Update PHPCS versions to stable ones

adamziel:


rodrigoprimo:
@adamziel, I checked this PR and I have two questions. I don't know much about this project, so I might be missing some context.

- Is there a reason to downgrade PHPCompatibility from the `develop` branch to the `9.3.5` version? Usually, it is recommended to use the `develop` branch as `9.3.5` was released several years ago. That being said, in some cases, it is perfectly fine to use it.
- Why did you opt to exclude the `Tests` directory from the PHPCS checks?

adamziel:
> Is there a reason to downgrade PHPCompatibility from the develop branch to the 9.3.5 version?

No reason other than you've asked on Slack why we use the develop version and I thought maybe we should use the latest stable. If you recommend the `develop` branch, that works for me as well and I'm happy to restore it.

> Why did you opt to exclude the Tests directory from the PHPCS checks?

They follow different rules, e.g. camel cased filenames. It would be nice to define a set of rules that is useful for the test files later on, but for now I just wanted to keep moving.

rodrigoprimo:
Thanks, @adamziel.

> No reason other than you've asked on Slack why we use the develop version and I thought maybe we should use the latest stable. If you recommend the develop branch, that works for me as well and I'm happy to restore it.

On Slack, I was referring to PHPCS and not PHPCompatibility. I recommend to continue using the `develop` branch of PHPCompatibility.

> They follow different rules, e.g. camel cased filenames. It would be nice to define a set of rules that is useful for the test files later on, but for now I just wanted to keep moving.

That makes sense. If it is just the different rules for camel cased filenames or that and a few more, I would suggest just excluding the specific rules for the `Tests` directory. But keep moving and revisiting this later is fair as well.

This is what WordPress core does:

https://github.com/WordPress/wordpress-develop/blob/trunk/phpcs.xml.dist#L316-L318
https://github.com/WordPress/wordpress-develop/blob/trunk/phpcs.xml.dist#L274-L275

PR #156, opened 2025-09-05
Add SlevomatCodingStandard.ControlStructures.DisallowYodaComparison sniff

adamziel:
Adds a PHPCBF sniff to auto-fix yoda conditions that WordPress coding standards require

### DIFF
--- a/composer-ci-matrix-tests.json
+++ b/composer-ci-matrix-tests.json
@@ -25,7 +25,6 @@
     "require-dev": {
         "yoast/phpunit-polyfills": "^4.0.0",
         "phpcompatibility/php-compatibility": "^9.3.5",
-        "slevomat/coding-standard": "^8.21.1",
         "phpunit/phpunit": "8.5.x || ^9.5"
     },
     "autoload": {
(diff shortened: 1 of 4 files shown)


rodrigoprimo:
@adamziel, when I mentioned this sniff in our Slack conversation, I suggested it as something worth checking to see if it could be used once to automatically fix non-Yoda conditions in this effort to make this repository compatible with WPCS. I don't have previous experience with this sniff, and I don't know if what it flags and fixes match what WPCS flags.

I don't see a problem, per se, with including it in the repository. Just thought it was worth clarifying what I was suggesting.

adamziel:
@rodrigoprimo thank you for clarifying! It seems to be doing the right thing. We'll find out for sure when I do a PR to migrate the codebase – if it's not consistent with what WPCS needs, we'll just remove it. For me personally, Yoda conditions are unnatural to write and I'm always left with a few dozens of them to fix before merging anything – an automated sniff to do that is a huge quality of life improvement.

If there weren't other problems with that, I'd suggest an automated mirror repo that takes the original codebase and applies all the WPCS rules to it just to make the day-to-day easier, as in I'd like a minimal set of formatting rules similar to what prettier does in JS (tabs vs spaces, indentation, newlines, where curly braces are etc.) without the rest of it (yoda, filenames, snake_case, dots at the end of `// inline comments` etc.). Maybe one day 🤷

Issue #157, opened 2025-09-05
WPCS compliance


adamziel:
Using php-toolkit in WordPress core and core plugins (such as wordpress-importer) requires compliance with WordPress coding standards. Let's make this codebase compliant.

Related work:

* https://github.com/WordPress/php-toolkit/pull/156
* #155
* #153

cc @JanJakes

PR #168, opened 2025-09-05
PHPCS fixes

adamziel:
Another round of PHPCS fixes. This one disables some rules that were particularly impractical to swiftly resolve.

Let's follow-up with a CI job that checks the remaining rules.

Related to #157


adamziel:
A few tests needs adjusting and we're mostly PHPCS compliant!

PR #172, opened 2025-09-06
Add README to every component

adamziel:
🚧  WIP, these are all AI-generated, they need reading and rewriting before merging


adamziel:
The AI generated content is just too wrong. We'll need to write those by hand.

PR #174, opened 2025-09-06
GitHub workflow for package release and split

adamziel:
Adds a GitHub workflow for package release and split. This PR also:

* Cleans up inconsistencies in composer.json packages
* Removes the composer.json files nested in `vendor-patched` directories to avoid changing them when monorepo-builder runs


components/Markdown/vendor-patched/symfony/yaml/composer.json
```diff
@@ -1,41 +1,40 @@
...
+            "email": "fabien@symfony.com"
+        },
+        {
+            "name": "Symfony Community",
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "require": {
+        "php": ">=8.2",
+        "symfony/deprecation-contracts": "^0.0.11",
```
adamziel: Changing nested composer.json files is undesired and needs to be resolved before merging

PR #179, opened 2025-09-10
URLInTextProcessor: Test against the WHATWG URL test suite

adamziel:
> @dmsnell raised a good question in [https://github.com/WordPress/wordpress-importer/pull/202/](https://github.com/WordPress/wordpress-importer/pull/202/):
>
> > Have we audited against the WHATWG spec to see if we accept things we should reject, or reject things we should accept? Off the top of my head I don’t remember how auth in URLs is handled in the spec, but I doubt there’s a strong case for rewriting URLs with embedded auth.

**What this does**

This PR vets our URL text matcher against the WHATWG URL test suite and tightens the parsing to make smarter, pragmatic matches in real-world prose. The goal isn’t to be a full, spec-compliant parser; it’s to extract migration-relevant URLs from text with high precision and minimal false positives.

**Why not strictly WHATWG?**

The WHATWG parser assumes a base URL, a controlled environment, and full parse/serialize semantics. Inline detection in user content has different constraints: no base URL, lots of punctuation and whitespace noise, and a strong bias toward HTTP(S) links used for site migrations. The matcher therefore makes deliberate trade-offs.

## Deliberate deviations from WHATWG

**Reject URLs with embedded credentials**
`https://user:pass@wp.org` is not matched. Rewriting URLs that presume transferable credentials is hazardous and rarely correct for migrations.

**Reject non-HTTP(S) schemes**
Out of scope for site moves; all of these are rejected:
`gopher://site.com`, `blob:afgh2-48189d`, `ahttp://site.com`, `mailto:user@site.com`, `file://asset.zip`.
If we need additional schemes later, we can add them intentionally.

**Reject non‑absolute‑looking references**
While we do rely on a base URL, inputs like `::`, `/index.html`, `?query` are still ignored. Bare-domain forms like `mysite.org/?query` are still matched.

**Handle trailing punctuation sensibly**
`https://mysite.com/path/..` is interpreted as `https://mysite.com/path/` rather than collapsing to the origin. A final period is far more likely sentence punctuation than `../`. If a user truly writes `https://mysite.com/path/../`, we parse it as expected.

**Fuzzy matching for malformed ports**

* **WHATWG**: `"http://w.org:100000 plugins are in the directory" → failure`.
* **Inline detection**: `"http://w.org:100000 plugins are in the directory" → "http://wp.org/"` (truncate at the invalid port).

This is a best‑effort extraction of the valid prefix rather than an all‑or‑nothing rejection.

**Whitespace handling**

* **WHATWG**: `"http://example\t.\norg" → "http://example.org/"`.
* **Inline detection**: stops at the first whitespace, yielding `"http://example/"`.

This reflects how URLs actually appear in text blocks where whitespace often terminates a link.

## Tests

The WHATWG URL test suite is copied and adapted for “URL-in-text” matching, with additional cases covering credentials, schemes, punctuation, malformed ports, and whitespace. CI exercises the full set.


dmsnell:
> Fuzzy matching for malformed ports

We may want to reject these as well, given that the invalid port looks well-formed. Why? Because we know that the original URL _could not have pointed to a valid resource_, so we don’t have to try and rewrite it, given that we know it will still point to an invalid resource.

> This reflects how URLs actually appear in text blocks where whitespace often terminates a link.


Yes, and worth pointing out that the WHATWG parser and tests are attempting to validate entire strings which are potentially complete URLs. This is a major source of difference that I have found when trying to fulfill a purpose like you are doing, which is _determining which substrings inside a text are valid URLs_.

I would love to have a more standardized approach to this, but it may not exist.

adamziel:
Refined GitHub highlights a URL we'd reject :D 

CleanShot 2025-09-10 at 16 35 11@2x

adamziel:
> We may want to reject these as well, given that the invalid port looks well-formed. Why? Because we know that the original URL _could not have pointed to a valid resource_, so we don’t have to try and rewrite it, given that we know it will still point to an invalid resource.

I rephrased the example to:

* WHATWG: `"http://w.org:100000 plugins are in the directory"` → failure.
* Inline detection: `"http://w.org:100000 plugins are in the directory"` → "http://wp.org/" (truncate at the invalid port).

I wonder what's your take on that

components/DataLiberation/Tests/URLInTextProcessorTest.php
```diff
@@ -291,28 +291,6 @@ public static function provider_test_finds_next_url_when_base_url_is_used() {
...
-				'http://user%40name:pass@example.com',
-				'http://user%40name:pass@example.com/',
-				'Link: http://user%40name:pass@example.com',
-			),
-			'Password with Special Characters'            => array(
-				'http://user:pa%40ss@example.com',
-				'http://user:pa%40ss@example.com/',
-				'Secure link: http://user:pa%40ss@example.com',
-			),
-
```
dmsnell, line 315: we can move these to rejection tests to show that it’s supposed to not detect them, instead of removing the test cases

components/DataLiberation/Tests/URLInTextProcessorWHATWGComplianceTest.php
```diff
@@ -0,0 +1,106 @@
...
+	 */
+	public function test_whatwg_url_in_text( $example ) {
+		$input_url = $example['input'];
+		$base_url = $example['base'];
+		$should_be_valid = !isset( $example['failure'] ) || !$example['failure'];
+		
+		$processor = new URLInTextProcessor( "Visit $input_url for more info", $base_url );
+		if ( ! $should_be_valid ) {
+			if ( false === $processor->next_url() ) {
+				$this->assertTrue( true, "Should not have found URL in text: '$input_url'" );
```
dmsnell: I’m confused by these tests. Why are we going out of our way to mix booleans?

```php
if ( ! $should_be_valid ) {
	$this->assertFalse(
		$processor->next_url(),
		"Should not…"
	);
}
```

components/DataLiberation/Tests/URLInTextProcessorWHATWGComplianceTest.php
```diff
@@ -0,0 +1,106 @@
...
+				return;
+			}
+			$parsed_url = $processor->get_parsed_url();
+			if ( ! $parsed_url ) {
+				$this->assertTrue( true, "Should not have parsed URL: '$input_url'" );
```
dmsnell: same here, this feels very strange and is hard to follow — what am I missing when I look at this and wonder about the garden-path we’re taking to obscure the assertion?

components/DataLiberation/Tests/URLInTextProcessorWHATWGComplianceTest.php
```diff
@@ -0,0 +1,106 @@
...
+			if ( is_string( $example ) ) {
+				continue;
+			}
+			
+			$filtered_examples[] = array( $example );
+		}
+
+		return $filtered_examples;
+	}
+}
```
dmsnell, line 94: did you write all this code?

components/DataLiberation/URL/class-urlintextprocessor.php
```diff
@@ -124,9 +128,8 @@ public function __construct( $text, $base_url = null ) {
 		// Source: https://github.com/vstelmakh/url-highlight/blob/master/src/Matcher/Matcher.php.
 		$this->regex = '/' . $prefix . '
             (?:                                                      # scheme
-                (?<scheme>https?:)?                                  # Only consider http and https
-                \/\/                                                 # The protocol does not have to be there, but when
-                                                                     # it is, is must be followed by \/\/
+                (?<scheme>[a-z0-9\+]+?:)?                            #
+                (?:\/|\/\/)?                                         # The protocol may optionally be followed by one or two slashes
```
dmsnell: which supported URLs do we have where there’s only a single slash?

adamziel: WHATWG surprised me by how lenient it is on slashes:

```js
> (new URL("https://example.com/")).toString()
'https://example.com/'

> (new URL("https:/example.com/")).toString()
'https://example.com/'

> (new URL("https:example.com/")).toString()
'https://example.com/'
```

dmsnell: 🤨

Safari’s URL bar works with one, but diverts to the search engine if no slash is there.

This is annoying.

adamziel: it is!

components/DataLiberation/Tests/URLInTextProcessorTest.php
adamziel: good call!

components/DataLiberation/Tests/URLInTextProcessorWHATWGComplianceTest.php
adamziel: Ah, I was doing some debugging and forgot to clean this up.

components/DataLiberation/Tests/URLInTextProcessorWHATWGComplianceTest.php
adamziel: I copied parts of URLParserWHATWGComplianceTest, expanded the list of exclusions, then contracted it and I guess I never cleaned this up

components/DataLiberation/Tests/URLInTextProcessorWHATWGComplianceTest.php
adamziel: Done

components/DataLiberation/Tests/URLInTextProcessorWHATWGComplianceTest.php
adamziel: Cleaned up

components/DataLiberation/Tests/URLInTextProcessorWHATWGComplianceTest.php
adamziel: Cleaned up

components/DataLiberation/URL/class-urlintextprocessor.php
adamziel: I've adjusted it to match zero or more slashes after all to make it capture the entire `file:///path` URL and reject it as a whole (as opposed to matching just the `//path` subset and assuming it's a protocol-relative URL)

components/DataLiberation/Tests/URLInTextProcessorTest.php
adamziel: done

adamziel:
> > We may want to reject these as well, given that the invalid port looks well-formed. Why? Because we know that the original URL _could not have pointed to a valid resource_, so we don’t have to try and rewrite it, given that we know it will still point to an invalid resource.
> 
> I rephrased the example to:
> 
> * WHATWG: `"http://w.org:100000 plugins are in the directory"` → failure.
> * Inline detection: `"http://w.org:100000 plugins are in the directory"` → "[http://wp.org](http://wp.org/)" (truncate at the invalid port).
> 
> I wonder what's your take on that

I'll go ahead and merge this PR. Let's keep this conversation going separately. I'm happy to start rejecting those URLs if we'll conclude it's a better trade-off than not rejecting them.

PR #181, opened 2025-09-10
BlockMarkupUrlProcessor: Iterate over the attributes using names, not indices

adamziel:
@dmsnell [noted](https://github.com/WordPress/wordpress-importer/pull/202#discussion_r2334173254):

> the passing around of the index with the `-1` sentinel value is hard for me to follow. did you consider setting something like `$this->currently_examined_attribute_name = null;`?

This PR switches from to using indices to using a list of names.

> or is the issue reentrancy in all of these functions?

Block markup parser is not re-entrant. We should never need to process block markup that blows up the memory so I didn't bother implementing re-entrancy there.

# Testing instructions

CI

### DIFF
--- a/components/DataLiberation/BlockMarkup/class-blockmarkupurlprocessor.php
+++ b/components/DataLiberation/BlockMarkup/class-blockmarkupurlprocessor.php
@@ -22,7 +22,14 @@ class BlockMarkupUrlProcessor extends BlockMarkupProcessor {
 	private $base_url_object;
 	private $url_in_text_processor;
 	private $url_in_text_node_updated;
-	private $inspected_url_attribute_idx = - 1;
+	private $attributes_to_inspect;
+
+	/**
+	 * The name of the URL attribute whose value is currently being inspected.
+	 *
+	 * @var string|null
+	 */
+	private $currently_inspected_url_attribute_name;
 
 	public function __construct( $html, ?string $base_url_string = null ) {
 		parent::__construct( $html );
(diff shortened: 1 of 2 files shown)


components/DataLiberation/BlockMarkup/class-blockmarkupurlprocessor.php
```diff
@@ -116,39 +131,51 @@ private function next_url_in_text_node() {
...
+		} else {
+			/**
+			 * Forget the attribute we've inspected on the previous call to
+			 * next_url_attribute().
+			 */
+			array_pop( $this->inspecting_html_attributes );
+		}
 
+		while ( count( $this->inspecting_html_attributes ) > 0 ) {
+			$attr      = $this->inspecting_html_attributes[ count( $this->inspecting_html_attributes ) - 1 ];
```
dmsnell, line 155: I suppose the order does or doesn’t matter, and that would suggest `[count( $ ) - 1]` or `[0]` is more appropriate. `[0]` is clear and easy to reason about. `array_pop()` is nice because I think it mostly updates an internal counter and frees the last item, vs `array_shift()` which does a mem copy?

or maybe not. I skimmed the PHP source and it looks like `array_shift()` might just unset the element and update the array value’s internal pointer, so maybe they are comparable in performance.

adamziel: `array_shift()` seems to be much slower empirically, which is why I went for `count( $ ) - 1` even though I don't like it visually:

```
array_pop duration: 0.14569401741028 seconds
array_shift duration: 3.1403939723969 seconds
```

```php
<?php

function benchmark_array_operations($iterations = 10000) {
    $array = range(1, 1000);

    // Benchmark array_pop
    $start_time = microtime(true);
    for ($i = 0; $i < $iterations; $i++) {
        $temp_array = $array;
        while (!empty($temp_array)) {
            array_pop($temp_array);
        }
    }
    $pop_duration = microtime(true) - $start_time;

    // Benchmark array_shift
    $start_time = microtime(true);
    for ($i = 0; $i < $iterations; $i++) {
        $temp_array = $array;
        while (!empty($temp_array)) {
            array_shift($temp_array);
        }
    }
    $shift_duration = microtime(true) - $start_time;

    echo "array_pop duration: " . $pop_duration . " seconds\n";
    echo "array_shift duration: " . $shift_duration . " seconds\n";
}

benchmark_array_operations();
```

dmsnell: the thing that seemed different was that it could trigger re-hashing the array. if that’s the case, it would definitely over-represent in a synthetic microbenchmark.

it all seems reasonable what you are doing though 👍

PR #184, opened 2025-09-11
FileReadStream improvement

adamziel:
Address [this comment](https://github.com/WordPress/wordpress-importer/pull/202/#discussion_r2334500123) from @brandonpayton by moving the assignment after the check:

> If this seek fails, bytes_already_forgotten will be left at the invalid target offset. Is that desired?


brandonpayton (approved):
This looks like a good change.

It might be good to add a FileWriteStream test to test the result of failing to seek.

adamziel:
I still need to backport this one to https://github.com/WordPress/wordpress-importer/pull/202/

PR #188, opened 2025-09-17
Detect relative URLs in known block attributes

adamziel:
Gives `BlockMarkupUrlProcessor` knowledge about which block attributes are designed to contain a relative URL. 

Known URL attributes can be assumed to hold a URL and be parsed with the base URL. For example, a `/about-us` value in a `wp:navigation-link` block’s url attribute is a relative URL to the `/about-us` page.

Other attributes may or may not contain URLs, and we cannot assume they do. A value like `/about-us` could be a relative URL or a class name. In those cases we’ll ignore relative URLs and only detect absolute URLs to avoid treating every string as a URL; this requires parsing without a base URL.  

Related to https://github.com/WordPress/wordpress-playground/issues/1780

## Implementation details

This PR ships a list of all block attributes that are meant to hold a URL. It's similar to how the HTML spec declares a list of all [HTML attributes meant to hold a URL](https://html.spec.whatwg.org/multipage/indices.html#attributes-1).

It also ships a filter the extenders can use to nudge the URL rewriter to treat a block attribute as either:

```php
$is_relative_url_block_attribute = apply_filters(
	'url_processor_is_relative_url_block_attribute',
	$is_relative_url_block_attribute,
	array(
		'block_name' => $this->get_block_name(),
		'attribute_key' => $this->get_block_attribute_key(),
	)
)
```

A hypothetical plugin shipping a `wp:custom-image` block could nudge the URL parser to treat its `src` attribute as a relative URL by registering the following filter:

```php
<?php
function recognize_custom_image_block_attribute_as_relative_url ( $is_known, $context ) {
	if (
		'wp:custom-image' === $context['block_name'] &&
	     'src' === $context['attribute_key']
	) {
		return true;
	}

	return $is_known;
}

add_filter(
	'url_processor_is_relative_url_block_attribute',
	'recognize_custom_image_block_attribute_as_relative_url',
	10,
	2
);
```

### Other changes

This PR renames two constants for clarity:

* `URL_ATTRIBUTES` -> `HTML_ATTRIBUTES_TO_ACCEPT_RELATIVE_URLS_FROM`
* `URL_ATTRIBUTES_WITH_SUBSYNTAX` -> `HTML_ATTRIBUTES_WITH_SUBSYNTAX_TO_ACCEPT_RELATIVE_URLS_FROM`
* `URL_CONTAINING_TAGS_WITH_SUBSYNTAX` -> `HTML_TAGS_WITH_SUBSYNTAX_TO_ACCEPT_RELATIVE_URLS_FROM`

## Testing

CI – see the new tests shipped with this PR.

### DIFF
--- a/components/DataLiberation/BlockMarkup/class-blockmarkupurlprocessor.php
+++ b/components/DataLiberation/BlockMarkup/class-blockmarkupurlprocessor.php
@@ -419,7 +419,7 @@ class BlockMarkupUrlProcessor extends BlockMarkupProcessor {
 	 * A list of block attributes that are known to contain URLs.
 	 *
 	 * It covers WordPress core blocks as of WordPress version 6.9. It can be
-	 * extended by plugins and themes via the "wp_data_liberation_is_known_url_block_attribute"
+	 * extended by plugins and themes via the "url_processor_is_relative_url_block_attribute"
 	 * filter.
 	 *
 	 * @var array
--- a/components/DataLiberation/BlockMarkup/class-blockmarkupurlprocessor.php
+++ b/components/DataLiberation/BlockMarkup/class-blockmarkupurlprocessor.php
@@ -478,7 +478,7 @@ class BlockMarkupUrlProcessor extends BlockMarkupProcessor {
 	 * @TODO: Either explicitly support these attributes, or explicitly drop support for
 	 *        handling their subsyntax. A generic URL matcher might be good enough.
 	 */
-	public const HTML_ATTRIBUTES_TO_ACCEPT_RELATIVE_URLS_FROM_WITH_SUBSYNTAX = array(
+	public const HTML_ATTRIBUTES_WITH_SUBSYNTAX_TO_ACCEPT_RELATIVE_URLS_FROM = array(
 		'*'      => array( 'style' ), // background(), background-image().
 		'APPLET' => array( 'archive' ),
 		'IMG'    => array( 'srcset' ),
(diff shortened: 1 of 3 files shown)


components/DataLiberation/BlockMarkup/class-blockmarkupurlprocessor.php
```diff
@@ -362,6 +400,16 @@ public function get_inspected_attribute_name() {
 		return $this->inspecting_html_attributes[ count( $this->inspecting_html_attributes ) - 1 ];
 	}
 
+	public const URL_BLOCK_ATTRIBUTES = array(
```
adamziel: Let's audit it – maybe there's more

components/DataLiberation/BlockMarkup/class-blockmarkupurlprocessor.php
```diff
@@ -184,22 +184,60 @@ private function next_url_attribute() {
...
+			 * absolute URLs to avoid treating every string as a URL. This requires
+			 * parsing without a base URL.
 			 */
-			if ( is_string( $url_maybe ) ) {
+			$is_known_url_block_attribute = (
+				isset( self::URL_BLOCK_ATTRIBUTES[ $this->get_block_name() ] ) &&
+				in_array( $this->get_block_attribute_key(), self::URL_BLOCK_ATTRIBUTES[ $this->get_block_name() ], true )
+			);
+
+			$is_known_url_block_attribute = apply_filters(
```
adamziel: a docstring would help to surface this more for the extenders

PR #197, opened 2025-10-29
CSSTokenizer

adamziel:
## Overview

Ships a CSSTokenizer class that tokenizes CSS syntax. It follows the [CSS Syntax Level 3 spec](https://www.w3.org/TR/css-syntax-3/), giving Data Liberation tooling a reliable foundation for discovering and rewriting URLs without leaning on brittle regular expressions or partial parsers.

This branch is tested using the [`@rmenke/css-tokenizer-tests`](https://github.com/romainmenke/css-tokenizer-tests/) test corpus and an extra set of tests designed to cover nuances.

### On-the-fly normalization

The CSS Spec requires the following normalization step:

> Replace any U+000D CARRIAGE RETURN (CR) code points, U+000C FORM FEED (FF)
> code points, or pairs of U+000D CARRIAGE RETURN (CR) followed by U+000A LINE
> FEED (LF) in input by a single U+000A LINE FEED (LF) code point.
> Replace any U+0000 NULL or surrogate code points in input with U+FFFD REPLACEMENT
> CHARACTER (�).

This processor delays normalization as much as possible, rather than preprocessing
the entire input upfront. This avoids the upfront allocation cost for clean CSS
and preserves original byte positions for accurate raw token extraction. A part
of the normalization is performed on-the-fly as the tokens are consumed. The rest
of it is done once the token value is requested.

### No EOF token

The EOF token is a CSS parsing concept, not CSS tokenization concept. Therefore,
this processor does not produce it.

### UTF-8 handling

Only UTF-8 strings are supported. In case an invalid UTF-8 sequence is encountered,
it is replaced with a U+FFFD REPLACEMENT CHARACTER (�) using the maximal subpart
approach described in https://www.unicode.org/versions/Unicode9.0.0/ch03.pdf,
section 3.9 Best Practices for Using U+FFFD.

## Usage

The next_token() method is the main entry point for tokenizing a CSS string.
It will consume the next token from the input stream and return true if a token
was found. Otherwise, it will return false:

```php
$css = 'width: 10px;';
    echo $processor->get_normalized_token();
}
// Outputs:
// width: 10px;
```


adamziel:
@dmsnell I need to adjust the line endings in a few test examples, but other than that this is ready for a review. This PR is focused on tokenizing CSS without any URL rewriting logic yet. LLMs were involved and so was a lot of manual reviewing and rewriting.

components/DataLiberation/Tests/css-test-cases.php
```diff
@@ -0,0 +1,5390 @@
...
+ * CSS Tokenizer Test Cases generated from @rmenke/css-tokenizer-tests.
+ * Indices are UTF-8 byte offsets.
+ * String and URL values are already decoded.
+ */
+
+return array (
+  'tests/at-keyword/0001' => 
+  array (
+    'css' => '@foo
+',
```
dmsnell: is this auto-generated? If so, I have found it useful for the sake of readability to replace literal whitespace with Unicode escapes, making them ever-more explicit as being text under test.

e.g. `'css' => "@foo\n"`

components/DataLiberation/Tests/CSSTokenizerTest.php
```diff
@@ -0,0 +1,918 @@
...
+		// Check structured data (if present in expected)
+		// We'll do a loose comparison here since our implementation might differ slightly
+		if ( isset( $expected['value'] ) && null !== $expected['value'] ) {
+			$this->assertArrayHasKey( 'value', $actual, $message . ' (value missing)' );
+			// Loose comparison because floats may differ slightly
+			$this->assertEquals(
+				$expected['value'],
+				$actual['value'],
+				$message . ' (value)',
+				0.0001  // delta for float comparison
```
dmsnell: `assertEqualsWithDelta()` exists for this check.

though, perhaps we should be careful about converting to floats when lexing or parsing and prefer instead to retain the literal values from the CSS. there are some known and nasty issues with casting, including PHP automatically using locale-specific separators that break CSS if applied (e.g. CSS decimal separator is _always_ `.` even in Germany where `,` is typically used instead).

components/DataLiberation/Tests/CSSTokenizerTest.php
```diff
@@ -0,0 +1,918 @@
...
+		);
+
+		// Compare each token
+		foreach ( $expected_tokens as $index => $expected_token ) {
+			if ( ! isset( $actual_tokens[ $index ] ) ) {
+				$this->fail( "Missing token at index $index for CSS: " . var_export( $css, true ) );
+			}
+			$actual_token = $actual_tokens[ $index ];
+			$this->assert_token_matches( $expected_token, $actual_token, $index, $css );
+		}
```
dmsnell: out of curiosity, was something insufficient with `->assertSame()`? my expectation would be that it would show a useful debug message be default if we relied on it.

additionally there is `assertArrayIsEqualToArrayOnlyConsideringListOfKeys()` which can also be used with `array_keys( $expected_token )` if there are keys we want to ignore, but while still ensuring that an optional `value` key is checked.
```suggestion
		}
```

components/DataLiberation/Tests/CSSTokenizerTest.php
```diff
@@ -0,0 +1,918 @@
...
+			$test_cases = require __DIR__ . '/css-test-cases.php';
+		}
+
+		$data = array();
+		foreach ( $test_cases as $test_name => $test_case ) {
+			$data[ $test_name ] = array(
+				$test_case['css'],
+				$test_case['tokens'],
+			);
+		}
```
dmsnell: not that it matters, but we can also use `yield $test_name => array( $test_case['css'], $test_case['tokens'] )`

though this makes me wonder more, why are we doing this at all? why are we unpacking the generated test cases to reassemble them in the exact same shape as we chose to represent them in the first place?

did you make these choices?

components/DataLiberation/URL/class-csstokenizer.php
```diff
@@ -0,0 +1,1684 @@
...
+			$this->get_codepoint_at( $this->at, $matched_bytes );
+
+			// We're in trouble!
+			// If get_codepoint_at fails to advance, we're dealing with a non-UTF-8 sequence.
+			// @TODO: Decide what's the most useful strategy for handling this. Some form of emitting
+			// an error would be useful for sure.
+			// For now, we'll just consume that byte to prevent infinite loop and keep processing.
+			if ( 0 === $matched_bytes ) {
+				$matched_bytes = 1;
+			}
```
dmsnell, line 547: I tried various ways to verify this check and had trouble. if we use `_wp_scan_utf8()` then it will tell us how to map invalid bytes to invalid code points.

If we get invalid bytes then whatever that returns as the invalid length represents how many bytes are to be interpreted as a single replacement character, which we could use here. we should not advance by a single byte just because there’s an error. for instance, if we get a valid prefix of three out of the four bytes for a character, we are to treat all three of those as a single replacement, whereas if those three bytes are not a valid prefix, they should be split into sections that are. this is called the “maximal subpart” replacement and will match what new versions of `mb_scrub()` will return, or likewise `wp_scrub_utf8()`.

```php
$at = $this->at;
$invalid_length = 0;
if ( 1 !== _wp_scan_utf8( $matched_bytes, $at, $invalid_length, null, 1 ) ) {
	$matched_bytes = $invalid_length;
}
```

the displayed character here should be U+FFFD (�) but the bytes will be the invalid bytes.

components/DataLiberation/URL/class-csstokenizer.php
```diff
@@ -0,0 +1,1684 @@
...
+					$this->at = $save_pos;
+				}
+			}
+		}
+
+		// Convert string to a number, and set the value to the returned value.
+		// We use a PHP typecast as it's mostly compatible with the spec's behavior.
+		// @TODO: Investigate any differences.
+		// See https://www.w3.org/TR/css-syntax-3/#convert-a-string-to-a-number.
+		$this->token_value = (float) substr( $this->css, $start, $this->at - $start );
```
dmsnell: noted above in the tests file, I think this is a questionable choice. there are rare times when we want to perform computation on the CSS values, but usually we want to simply know how to rearrange them, add to them, or remove them.

to that end, automatic conversion to numbers is going to cause problems because PHP numbers are different from JavaScript numbers and those are both different than CSS numbers.

if we leave these as a string we acknowledge that the literals in the CSS are distinct types and we will guarantee preservation of them. otherwise, by converting here, we encourage auto-conversion and the changing of data due to a failure to round-trip.

components/DataLiberation/URL/class-csstokenizer.php
```diff
@@ -0,0 +1,1684 @@
...
+			// Convert the hex digits to a UTF-8 string.
+			return codepoint_to_utf8_bytes( hexdec( $hex ) );
+		}
+
+		// Anything else.
+		// Return the current input code point.
+		// Null bytes are replaced with U+FFFD during preprocessing.
+		if ( "\x00" === $this->css[ $at ] ) {
+			$bytes_consumed = 1;
+			return "\xEF\xBF\xBD"; // U+FFFD.
```
dmsnell: why not use `"\u{FFFD}"`?

dmsnell:
Overall, the organization is clear and I like how this is similar to the HTML API interfaces. Should be easy to use to find and modify a URL, for example.

components/DataLiberation/URL/class-csstokenizer.php
adamziel: ooo I forgot that existed, thank you!

components/DataLiberation/URL/class-csstokenizer.php
adamziel: Good point, I've removed this. If anyone will want to actually process these numeric values, we'll discuss then.

components/DataLiberation/Tests/CSSTokenizerTest.php
adamziel: Good spot! This method came from Claude. I've simplified it a bit but didn't notice it's just empty cycles recreating the same data structure. I'll replace it with `return require "examples.php"`.

components/DataLiberation/Tests/css-test-cases.php
adamziel: Yes, this was auto-generated with a script – these are exactly the line endings I've meant in my last comment. I've updated it to use escape sequences (`\n`) instead of literal codepoints (newlines, carriage returns etc).

components/DataLiberation/Tests/CSSTokenizerTest.php
adamziel: Good idea, I've migrated to assertSame 👍

components/DataLiberation/Tests/CSSTokenizerTest.php
adamziel: Thanks and TIL about assertEqualsWithDelta. Since we're no longer parsing floats, we don't the delta either – I've just removed this.

components/DataLiberation/URL/class-csstokenizer.php
adamziel: I'm learning so much here, thank you for sharing this. I'll use _wp_scan_utf8 instead.

adamziel: I've brought over the new functions without replacing the old ones yet to keep this PR simpler. I'll migrate the rest of the repo to the new utilities in a follow-up.

adamziel:
I've applied your feedback @dmsnell, thank you! This PR seems in a pretty good place now. I'll draft a second one with set_token_value() before merging this, just to make sure we have the right APIs.

adamziel:
I think we're good API-wise. Even if not, we can always adjust since it's the php-toolkit repo.

adamziel:
CC @sirreal – you may like this

components/Blueprints/class-runner.php
```diff
@@ -383,7 +383,7 @@ private function load_blueprint() {
 		if ( function_exists( 'mb_check_encoding' ) ) {
 			$is_valid_utf8 = mb_check_encoding( $blueprint_string, 'UTF-8' );
 		} else {
-			$is_valid_utf8 = utf8_is_valid_byte_stream( $blueprint_string );
+			$is_valid_utf8 = ! _wp_has_noncharacters_fallback( $blueprint_string );
```
dmsnell, line 386: note too, @adamziel that this is a semantic change. you might have wanted that, but `wp_has_noncharacters()` specifically indicates if a string contains Unicode noncharacters, which are mostly (but not exclusively) the last two code points in each plane (U+xxFFFE, U+xxFFFF). it may return `true` if there is invalid UTF-8.

`wp_is_valid_utf8()` indicates if the byte stream is valid UTF-8. it may return `true` if there are noncharacters.

adamziel: Thanks! I've revisited it in https://github.com/WordPress/php-toolkit/pull/200

PR #198, opened 2025-10-30
CSSProcessor::set_token_value( $new_value )

adamziel:
Follows up on https://github.com/WordPress/php-toolkit/pull/197 by adding a `set_token_value( $new_value )` method to allow rewriting CSS. At the moment, it only supports updating the URL token value.

```php
$css = 'background: url(old.jpg);';

	if ( CSSProcessor::TOKEN_URL === $processor->get_token_type() ) {
		// URL with safe characters: letters, digits, hyphens, underscores, dots, slashes.
		$processor->set_token_value( "image😀.jpg ("special")" );
	}
}

echo $processor->get_updated_css();
// background: url("image😀.jpg (\22 special\22 )");
```

## Implementation details

`set_token_value( $new_value )` always uses the quoted URL syntax to encode the new URL. It only escapes quotes newline characters, backslashes, and double quotes. All other bytes are preserved as-is.

cc @dmsnell


components/DataLiberation/URL/class-cssprocessor.php
```diff
@@ -741,6 +787,165 @@ public function get_token_value_length(): ?int {
...
+			$this->lexical_updates,
+			function ( $a, $b ) {
+				return $b['start'] - $a['start'];
+			}
+		);
+
+		$css = $this->css;
+
+		foreach ( $this->lexical_updates as $update ) {
+			$css = substr_replace(
```
adamziel: This smells like a lot of large allocations and work. But maybe not? Let's benchmark that against stitching the string like the HTML tag processor does.

adamziel: I've compared substr_replace vs implode vs concatenating strings, and concatenating seems to be the fastest

PR #199, opened 2025-10-31
CSSProcessor::set_token_value( $new_value )

adamziel:
Follows up on https://github.com/WordPress/php-toolkit/pull/197 by adding a `set_token_value( $new_value )` method to allow rewriting CSS. At the moment, it only supports updating the URL token value. 

```php
$css = 'background: url(old.jpg);';

	if ( CSSProcessor::TOKEN_URL === $processor->get_token_type() ) {
		// URL with safe characters: letters, digits, hyphens, underscores, dots, slashes.
		$processor->set_token_value( "image😀.jpg ("special")" );
	}
}

echo $processor->get_updated_css();
// background: url("image😀.jpg (\22 special\22 )");
```

## Implementation details

`set_token_value( $new_value )` always uses the quoted URL syntax to encode the new URL. It only escapes quotes newline characters, backslashes, and double quotes. All other bytes are preserved as-is.

PR #198 was supposed to merge this into trunk, but I never updated the base.

cc @dmsnell


components/Blueprints/class-runner.php
```diff
@@ -383,7 +383,7 @@ private function load_blueprint() {
 		if ( function_exists( 'mb_check_encoding' ) ) {
 			$is_valid_utf8 = mb_check_encoding( $blueprint_string, 'UTF-8' );
 		} else {
-			$is_valid_utf8 = utf8_is_valid_byte_stream( $blueprint_string );
+			$is_valid_utf8 = ! _wp_has_noncharacters_fallback( $blueprint_string );
```
dmsnell, line 386: `wp_has_noncharacters()` from `utf8.php` is indeed much faster. unless there is a reason you need them, I would suggest sticking to the non-private methods if you can.

Issue #196, opened 2025-10-23
Migrate to the latest version of the UTF-8 API


adamziel:
@dmsnell significantly improved the speed of the UTF-8 decoder, let's bring those improvements into this project.

dmsnell:
https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/compat-utf8.php

https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/utf8.php

adamziel:
Resolved in #200 and #203

PR #200, opened 2025-10-31
Use wp_is_valid_utf8() and wp_scrub_utf8() from the new utf8.php decoder

adamziel:
Replaces two instances of the old UTF-8 decoding utilities with the new utf-8.php toolkit by @dmsnell:

* https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/compat-utf8.php
* https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/utf8.php

This PR only touches two tactical usages of the old tools:

* Blueprint validation now uses `wp_is_valid_utf8`
* CSSProcessor now uses `wp_scrub_utf8` instead of `_wp_scrub_utf8_fallback`

More refactoring is coming once there's a faster alternative to `_wp_scan_utf8`, see https://core.trac.wordpress.org/ticket/63863#comment:51

Related to #196.

Follows up on https://github.com/WordPress/php-toolkit/pull/199 and https://github.com/WordPress/php-toolkit/pull/197.

## Testing instructions

If the CI passes, we're good. Unicode-related scenarios are covered by tests.


components/Blueprints/class-runner.php
```diff
@@ -379,12 +379,7 @@ private function load_blueprint() {
 		// Validate the Blueprint string we've just loaded.
 
 		// **UTF-8 Encoding:** Assert the Blueprint input is UTF-8 encoded.
-		$is_valid_utf8 = false;
-		if ( function_exists( 'mb_check_encoding' ) ) {
-			$is_valid_utf8 = mb_check_encoding( $blueprint_string, 'UTF-8' );
-		} else {
-			$is_valid_utf8 = ! _wp_has_noncharacters_fallback( $blueprint_string );
-		}
+		$is_valid_utf8 = ! wp_has_noncharacters( $blueprint_string );
```
dmsnell: I left a note in the other PR making this change, but it’s possible there is a misunderstanding here between _valid UTF-8_ and _noncharacters_, as noncharacters are valid UTF-8, and invalid UTF-8 cannot form noncharacters.

there is, however, `wp_is_valid_utf8()`

adamziel: Ah good point, I mixed it all up. Thank you for catching this.

adamziel:
`_wp_scan_utf8` became a major performance bottleneck in CI with 20x slowdown – the jobs were all timing out. I've explained the details in the UTF-8 trac ticket https://core.trac.wordpress.org/ticket/63863#comment:51. More involved migration to the new `wp_` functions is blocked until there's a way to keep the UTF-8 decoding speed closer to `utf8_codepoint_at`. I'll focus this PR on bringing over the regular `utf8.php` file and using the public API instead of private fallbacks.

PR #201, opened 2025-11-01
[XMLProcessor] ASCII fast path for parsing names

adamziel:
Speeds up XMLProcessor by consuming any ASCII bytes with `strspn` and avoiding calls to the utf8 decoder for most tags out there. The PHPUnit test suite for WXR files It speeds up parsing the 10MB WXR file in the test set from ~1.7s on average to ~0.6s on average.

This PR also moves from `utf8_codepoint_at` to `_wp_scan_utf8` for UTF-8 decoding without any speed penalty – see https://github.com/WordPress/php-toolkit/pull/200 for prior context.

cc @dmsnell

### DIFF
--- a/components/XML/class-xmlprocessor.php
+++ b/components/XML/class-xmlprocessor.php
@@ -18,11 +18,11 @@ use function WordPress\Encoding\utf8_ord;
  * It implements a subset of the XML 1.0 specification (https://www.w3.org/TR/xml/)
  * and supports XML documents with the following characteristics:
  *
- * * XML 1.0
- * * Well-formed
- * * UTF-8 encoded
- * * Not standalone (so can use external entities)
- * * No DTD, DOCTYPE, ATTLIST, ENTITY, or conditional sections (will fail on them)
+ * – XML 1.0
+ * – Well-formed
+ * – UTF-8 encoded
+ * – Not standalone (so can use external entities)
+ * – No DTD, DOCTYPE, ATTLIST, ENTITY, or conditional sections (will fail on them)
  *
  * XML 1.1 is explicitly not a design goal here. Version 1.1 is
  * more complex specification and not so widely supported.
@@ -1196,8 +1196,8 @@ class XMLProcessor {
 			/**
 			 * Compute fully qualified attributes and assert:
 			 *
-			 * * All attributes have valid namespaces.
-			 * * No two attributes have the same (local name, namespace) pair.
+			 * – All attributes have valid namespaces.
+			 * – No two attributes have the same (local name, namespace) pair.
 			 *
 			 * @see https://www.w3.org/TR/2006/REC-xml-names11-20060816/#uniqAttrs
 			 */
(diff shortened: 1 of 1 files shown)


components/XML/class-xmlprocessor.php
```diff
@@ -2328,17 +2326,25 @@ private function skip_whitespace() {
...
 			 * Parse the next unicode codepoint.
 			 *
-			 * We use a custom UTF-8 decoder here. No other method
-			 * is reliable and available enough to depend on it in
-			 * WordPress core:
+			 * We use a the `_wp_scan_utf8` UTF-8 decoder introduced in WordPress 6.9. No other method
+			 * is reliable and available enough to depend on it in WordPress core:
 			 *
-			 * * mb_ord() – is not available on all hosts.
+			 * * mb_ord() – is only available on 95% of hosts.
```
dmsnell: where is this number coming from? the last survey I saw, from _many_ years ago, showed 0.5% of hosts missing the `mbstring` extension

components/XML/class-xmlprocessor.php
```diff
@@ -2348,28 +2354,72 @@ private function parse_name( $offset ) {
...
-			 * Performance-wise, character-by-character processing via utf8_codepoint_at
-			 * is still much faster than relying on preg_match(). The mbstring extension
-			 * is likely faster. It would be interesting to evaluate the performance
-			 * and prefer mbstring whenever it's available.
+			 * Performance-wise, character-by-character processing via _wp_scan_utf8
+			 * is pretty slow. The ASCII fast path below enables skipping most of the
+			 * UTF-8 decoder calls.
+			 *
+			 * If the UTF-8 decoder performance ever becomes a bottleneck, there is a
+			 * few ways to significantly improve it:
```
dmsnell, line 2362: typo: “there _are_ a few ways”

components/XML/class-xmlprocessor.php
```diff
@@ -2348,28 +2354,72 @@ private function parse_name( $offset ) {
...
+			 *
+			 * * Call a native grapheme_ function when available.
+			 * * Introduce a custom UTF-8 decoder optimized for codepoint-by-codepoint processing.
+			 *   It could be the streaming version of the UTF-8 decoder, such as `_wp_iterate_utf8`,
+			 *   that avoids the repeated strspn() calls. Alternatively, the older `utf8_codepoint_at`
+			 *   function could be restored if its codepoint-by-codepoint decoding performance is
+			 *   better than the _wp_scan_utf8.
```
dmsnell, line 2369: might just be me, but the use of the `*` here for bullets looks more like a failed copy/paste than a bullet. I see this style is used more frequently than ` * - Item` throughout this code base so best not even addressing here, but they stood out to me

components/XML/class-xmlprocessor.php
adamziel: Ah, I've meant to quote that 0.5% figure but got the decimal point wrong :-)

components/XML/class-xmlprocessor.php
adamziel: Sure, let me adjust it

components/XML/class-xmlprocessor.php
dmsnell: it’s also probably overestimating the number of sites lacking `mbstring`, but it’s just the last number I have actually seen

PR #213, opened 2025-12-04
[JSON Validator] Tuple validation, multi-type enum checks

ashfame:
This PR addresses two related bugs in the schema validator that caused it to throw fatal errors (`UnsupportedSchemaException` or a `TypeError`) when it should have gracefully returned a `ValidationError`.

### The Problem

The validator would crash when attempting to validate data against a schema that used either of the following valid JSON Schema constructs:

1.  **Tuple Validation:** When an array's `items` keyword was an array of schemas (e.g., `{"type": "array", "items": [{"type": "string"}, {"type": "object"}]}`), the validator would throw an `UnsupportedSchemaException` instead of validating the data.
2.  **Multi-type Enums:** When a schema used an array for the `type` keyword along with an `enum` (e.g., `{"type": ["string", "number"]}`), the validator's internal integrity check would cause a fatal `TypeError`.

This meant that schemas that are perfectly valid according to the JSON Schema spec were not supported and would break the validation process entirely.

### Root Cause Analysis

1.  The `validate_array()` method was only implemented to handle "list validation" (where `items` is a single schema object). It lacked the logic to handle "tuple validation," causing it to misinterpret the array of schemas and throw an exception.
2.  The schema integrity check inside `validate_type()` was incorrectly calling `type_matches()`—which only accepts a string for the type—instead of the `type_matches_any()` method, which is designed to handle an array of types.

### The Solution

This PR implements two fixes to make the validator more compliant with the JSON Schema spec:

1.  **Added Tuple Validation Logic:** The `validate_array()` method now detects when the `items` keyword is an array (a tuple) and applies the correct validation logic, iterating through the data and schema arrays in lockstep.
2.  **Corrected the Enum Integrity Check:** The check in `validate_type()` has been updated to use the `type_matches_any()` method, allowing it to correctly validate schemas that use multiple types.

### How to Verify

I have added a new unit test, `testItReturnsValidationErrorForInvalidTupleData()`, which uses a schema with tuple validation.

*   **Before this fix:** Running this test causes the validator to throw a fatal `UnsupportedSchemaException`.
*   **After this fix:** The test now passes, as the validator correctly identifies the invalid data and returns a `ValidationError` object as expected.

### DIFF
--- a/components/Blueprints/Validator/class-humanfriendlyschemavalidator.php
+++ b/components/Blueprints/Validator/class-humanfriendlyschemavalidator.php
@@ -731,26 +731,34 @@ final class HumanFriendlySchemaValidator {
 	private function validate_array( array $path, array $data, array $schema ): ?ValidationError {
 		$children_errors = array();
 		if ( isset( $schema['items'] ) ) {
-			// In PHP, an associative array like {"type":"string"} is still an array.
-			// We need to distinguish between a list of schemas (tuple validation)
-			// and a single schema object (list validation). A simple way is to
-			// check if the keys are sequential integers from 0.
-			$is_tuple_validation = is_array( $schema['items'] ) && array_keys( $schema['items'] ) === range( 0, count( $schema['items'] ) - 1 );
+			// JSON schema supports two types of array validation:
+			//
+			// 1. List validation:
+			// `{"type":"string"}`
+			// A single schema object that all items in the data array must conform to.
+			// In PHP, this is represented as an associative array,
+			// `["type"=>"string"]`
+			//
+			// 2. Tuple validation:
+			// `[{"type":"string"},{"type":"object"}]`,
+			// A list of schemas, where each item in the data array is validated
+			// against the schema at the same index.
+			// In PHP, this is a numerically indexed array.
+			// `[0=>['type'=>'string'],1=>['type'=>'object']]`.
+			$is_tuple_schema = is_array( $schema['items'] ) && array_keys( $schema['items'] ) === range( 0, count( $schema['items'] ) - 1 );
 
-			if ( $is_tuple_validation ) {
-				// === TUPLE VALIDATION ===
-				// The validator doesn't support `additionalItems`, so we can report an error
-				// if the data array has more items than the schema tuple defines.
-				if ( isset( $schema['maxItems'] ) && count( $data ) > $schema['maxItems'] ) {
-					// This check is redundant if maxItems is always equal to count(items), but kept for robustness.
-				} elseif ( count( $data ) > count( $schema['items'] ) ) {
-					// This validator does not support `additionalItems`, so we treat extra items as an error.
+			if ( $is_tuple_schema ) {
+				// Tuple validation.
+				// Note: We do not support the "additionalItems" keyword.
+				// Therefore, we treat it as an error if the data array contains more items
+				// than are defined in the schema's "items" tuple.
+				if ( count( $data ) > count( $schema['items'] ) ) {
 					$children_errors[] = new ValidationError(
 						$this->convert_path_to_string( $path ),
 						'additional-items-not-allowed',
-						'Tuple validation failed: expected ' . count( $schema['items'] ) . ' items but got ' . count( $data ) . '.',
+						'Tuple validation failed: expected at most ' . count( $schema['items'] ) . ' items but got ' . count( $data ) . '.',
 						array(
-							'expectedCount' => count( $schema['items'] ),
+							'expectedMaxCount' => count( $schema['items'] ),
 							'actualCount' => count( $data ),
 						)
 					);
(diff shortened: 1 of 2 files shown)


ashfame:
@adamziel I purposefully pushed the failing test first so that you can see the exact test I added failing in CI. Second commit adds the fix, which makes all tests pass again (except 1 that's stuck in setting up PHP in CI).

components/Blueprints/Validator/class-humanfriendlyschemavalidator.php
```diff
@@ -731,10 +731,48 @@ private function validate_object( array $path, $data, array $schema ): ?Validati
...
 		$children_errors = array();
 		if ( isset( $schema['items'] ) ) {
-			foreach ( $data as $idx => $item ) {
-				$error = $this->validate_node( array_merge( $path, array( $idx ) ), $item, $schema['items'] );
-				if ( $error ) {
-					$children_errors[] = $error;
+			// In PHP, an associative array like {"type":"string"} is still an array.
+			// We need to distinguish between a list of schemas (tuple validation)
+			// and a single schema object (list validation). A simple way is to
+			// check if the keys are sequential integers from 0.
```
adamziel: looks like a good candidate for [array_is_list](https://php.net/array_is_list) polyfill (not a blocker here)

components/Blueprints/Validator/class-humanfriendlyschemavalidator.php
```diff
@@ -731,10 +731,48 @@ private function validate_object( array $path, $data, array $schema ): ?Validati
...
+						// This item has a specific schema defined in the tuple.
+						$error = $this->validate_node( array_merge( $path, array( $idx ) ), $item, $schema['items'][ $idx ] );
+						if ( $error ) {
+							$children_errors[] = $error;
+						}
+					}
+					// If there's no schema at $schema['items'][$idx], it's an "additional item", handled above.
+				}
+			} else {
+				// === LIST VALIDATION (existing logic) ===
```
adamziel: LLMs telltale sign: marking a code branch as "existing logic". Once we merge this PR, both branches become existing logic and that comment no longer makes sense. How about:

```suggestion
				// List validation
```

components/Blueprints/Validator/class-humanfriendlyschemavalidator.php
```diff
@@ -731,10 +731,48 @@ private function validate_object( array $path, $data, array $schema ): ?Validati
...
-				if ( $error ) {
-					$children_errors[] = $error;
+			// In PHP, an associative array like {"type":"string"} is still an array.
+			// We need to distinguish between a list of schemas (tuple validation)
+			// and a single schema object (list validation). A simple way is to
+			// check if the keys are sequential integers from 0.
+			$is_tuple_validation = is_array( $schema['items'] ) && array_keys( $schema['items'] ) === range( 0, count( $schema['items'] ) - 1 );
+
+			if ( $is_tuple_validation ) {
+				// === TUPLE VALIDATION ===
```
adamziel:
```suggestion
				// Tuple validation
```

components/Blueprints/Validator/class-humanfriendlyschemavalidator.php
adamziel: I'm still a bit confused about this logic. It's, effectively:

```php
$is_tuple_validation =   array_is_list( $schema['items'] );
$is_list_validation  = ! array_is_list( $schema['items'] );
```

Which reads like a contradiction. Also, a "list of schemas (tuple validation)" vs "single schema object (list validation)" confuses me, too. Can we find clearer terms to describe it?

I've looked it up in [the JSON schema spec](https://json-schema.org/understanding-json-schema/reference/array) and perhaps this quote would help:

```
* List validation: a sequence of arbitrary length where each item matches the same schema
* Tuple validation: a sequence of fixed length where each item may have a different schema. In this usage, the index (or location) of each item is meaningful as to how the value is interpreted. (This usage is often given a whole separate type in some programming languages, such as Python's tuple).
```

Although it tells me that `array_is_list()` is not the right check here, since we need to confirm we have a specific number of items, each defining a separate schema. It seems like this check needs to be more involved than it is now.

components/Blueprints/Validator/class-humanfriendlyschemavalidator.php
ashfame: Very true!

components/Blueprints/Validator/class-humanfriendlyschemavalidator.php
ashfame: @adamziel Thanks for the sharp & helpful feedback! I pushed another commit in hopes of making it more readable. Since "list" and "tuple" is how JSON schema spec defined it, I renamed the flag vars to `is_tuple_schema` after added a detailed explanation in the comments.

I have added a second check inside the tuple validation loop to make it a two-layered approach (first checking the schema's intent, then its integrity) which should fully address the concerns you raised. If I misunderstood, please let me know.

I have added some tuple related to test cases to the data provider of `testArrayValidation`.

ashfame: And better polyfills available on [array_is_list() page](https://www.php.net/array_is_list). Top voted note has the best one.

Do you want that one added to `components/Polyfill/php-functions.php` and used here instead?

components/Blueprints/Validator/class-humanfriendlyschemavalidator.php
```diff
@@ -731,10 +731,62 @@ private function validate_object( array $path, $data, array $schema ): ?Validati
...
+			// `[{"type":"string"},{"type":"object"}]`,
+			// A list of schemas, where each item in the data array is validated
+			// against the schema at the same index.
+			// In PHP, this is a numerically indexed array.
+			// `[0=>['type'=>'string'],1=>['type'=>'object']]`.
+			$is_tuple_schema = is_array( $schema['items'] ) && array_keys( $schema['items'] ) === range( 0, count( $schema['items'] ) - 1 );
+
+			if ( $is_tuple_schema ) {
+				// Tuple validation.
+				// Note: We do not support the "additionalItems" keyword.
```
adamziel, line 752: It would be cool to support it eventually. Not a blocker here

components/Blueprints/Validator/class-humanfriendlyschemavalidator.php
adamziel: @ashfame up to you, it doesn't seem like a big deal and maybe it's for the best that the validator doesn't have an implicit dependency on polyfills 🤔

adamziel:
Thank you @ashfame!

PR #219, opened 2026-03-12
CSS Processor: Fix consume_ident_start_codepoint bounds check

sirreal:
Fixes out-of-bounds string offset access when the input string ends while consuming an ident start.

### DIFF
--- a/components/DataLiberation/Tests/CSSProcessorTest.php
+++ b/components/DataLiberation/Tests/CSSProcessorTest.php
@@ -1530,5 +1530,15 @@ CSS;
 		$this->assertSame( "background: url(\"\xC0.jpg\");", $updated );
 	}
 
-
+	/**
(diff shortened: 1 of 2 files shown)


components/DataLiberation/Tests/CSSProcessorTest.php
```diff
@@ -1530,5 +1530,15 @@ public function test_set_token_with_invalid_utf8_sequence(): void {
 		$this->assertSame( "background: url(\"\xC0.jpg\");", $updated );
 	}
 
-
+	/**
+	 * Test bounds check when consuming and ident start token.
+	 */
+	public function test_trailing_hyphen_at_end_of_input(): void {
```
sirreal:
```suggestion
	public function test_ident_start_codepoint_bounds_check(): void {
```

adamziel:
Good catch!

PR #251, opened 2026-04-30
Smoke-test published Packagist packages in CI

adamziel:
The monorepo test suite cannot see how a published `wp-php-toolkit/*` package behaves in a downstream project, because the root `composer.json` autoloads every component through one big classmap that masks per-package autoload bugs. We have shipped three classes of break-the-consumer regressions in the past month alone:

- `autoload.files` pointing at a path that does not exist post-split (v0.7.0 zip);
- sibling-component `require_once` written for the monorepo `components/` layout (v0.7.0 polyfill);
- PSR-4 declarations against WordPress `class-*.php` filenames so the classes never resolve in a fresh consumer install (v0.7.1 zip, git, cli, filesystem, http-client, http-server, merge).

This PR closes that gap with a minimal smoke harness. For each of the 17 packages the script makes a scratch directory, runs `composer require wp-php-toolkit/<name>:<version>`, then walks every class, interface, and trait declared in the downloaded files (using PHP's tokenizer, so we never execute the package) and asserts the autoloader can reach each one. Anything unreachable fails the build with a list of the offending FQNs.

Two workflow hooks:

- `publish.yml` gains a `smoke-published-packages` job that runs after `release-composer-packages`, so a broken release is caught immediately, before the phar build even fires.
- `smoke-published.yml` runs nightly at 04:17 UTC against the latest tag as a continuous health check, with a `workflow_dispatch` input to re-run against any version on demand.

Validated locally against the live v0.7.1 release: the script correctly flags the seven PSR-4 packages that PR #250 fixed (cli, filesystem, git, http-client, http-server, zip, merge) and passes on blockparser, bytestream, html, polyfill, xml — exactly what we already knew about that release.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -101,6 +101,32 @@ jobs:
           git config --global user.name "github-actions[bot]"
           bash bin/split-code.sh
 
+  smoke-published-packages:
+    needs: release-composer-packages
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.2'
+
+      - name: Determine release tag
+        id: tag
+        run: |
+          git fetch --tags --prune --quiet || true
+          latest="$(git tag --list 'v[0-9]*' --sort=-version:refname | head -n 1)"
+          if [ -z "$latest" ]; then echo "No version tags found" >&2; exit 1; fi
+          echo "version=${latest#v}" >> "$GITHUB_OUTPUT"
+
+      - name: Smoke-test every published package
+        run: bash bin/smoke-published-packages.sh "${{ steps.tag.outputs.version }}"
+
   build-phar-files:
     needs: release-composer-packages
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


adamziel:
Pushed `ec83e431` adding a second smoke check (`bin/check-wp-coexistence.php`) that catches the regression seen in Automattic/jetpack#48424:

```
PHP Fatal error: Cannot declare class WP_Token_Map, because the name is already in use
  in /tmp/wordpress-latest/src/wp-includes/class-wp-token-map.php on line 146
```

Root cause was on us: in v0.7.1 I moved `html5-named-character-references.php` into HTML's `autoload.files` so the package would be self-contained. That bootstrap file calls `WP_Token_Map::from_precomputed_table(...)`, which dragged our `WP_Token_Map` into every consumer's process at composer-bootstrap time. The moment WordPress later loaded `wp-includes/class-wp-token-map.php` (no guard), PHP fatals on the second declaration.

The same commit fixes the underlying bug. The HTML package no longer eager-loads its char-references table; `WP_HTML_Decoder::read_character_reference` lazy-loads it on first call. In a WordPress runtime the global is initialized by core before our decoder ever runs, so we never include our copy. In a non-WP runtime the lazy path runs once and behaves exactly as before.

Reproduced the original break against the live `wp-php-toolkit/html:0.7.1`:

```
$ php bin/check-wp-coexistence.php
FAIL WordPress-coexistence: autoload.php eagerly declared WP-core classes:
  - WP_Token_Map
These will fatal when WordPress core later loads its own copy.
```

Monorepo `composer test` is still at parity with trunk (1 pre-existing failure, 0 errors).

PR #253, opened 2026-05-02
ci: enforce every docs snippet — no silent skips

adamziel:
## Summary
Tightens `bin/run-snippets.py` so every snippet in the docs catalog is verified, closing two gaps that let a broken snippet ship today:

- A **runnable** snippet that errored at runtime was dropped into a "couldn't run locally" bucket and the run still exited 0.
- A snippet flagged **non-runnable** (httpserver listener, cors-proxy daemon, etc.) was never executed *or* syntax-checked.

After this change:

| Catalog flag | What runs | What fails CI |
| --- | --- | --- |
| `runnable=True` | `php -d display_errors=stderr <snippet>` and stdout is normalized + compared to `_expected_outputs.json` (unless the snippet is in the `NO_EXPECTED` allowlist for unstable network/timing output) | Runtime failure, missing baseline, output drift |
| `runnable=False` | `php -l <snippet>` | Syntax errors |

There is no longer a silent-skip path. Failure output distinguishes syntax errors from runtime errors so logs point at the cause.

## Verification
- `bin/run-snippets.py --check` locally: **95 snippets pass** (87 runnable + output-verified, 8 non-runnable + syntax-verified).
- Fault-injection test (a runtime failure, a parse error in a non-runnable snippet, a new snippet without a baseline) — all three exit 1 with clear messages.
- The `snippet-tests.yml` workflow already runs `bin/run-snippets.py --check`, so no workflow change is needed beyond a refreshed header comment describing the new contract.

## Follow-up
**PR C:** make per-component `README.md` (or sibling `docs.md`) the source of truth for snippet code & metadata, instead of `bin/_docs_components.py` holding them as Python string literals.

## Test plan
- [ ] CI passes — `Verify docs snippets` runs against the new logic.
- [ ] If a future PR introduces a broken snippet, that PR fails CI loudly (you'll see this PR's enforcement working as designed once the next snippet PR comes through).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/snippet-tests.yml
+++ b/.github/workflows/snippet-tests.yml
@@ -1,12 +1,21 @@
 name: Verify docs snippets
 
-# Runs every PHP snippet in bin/_docs_components.py against the local toolkit
-# and compares stdout against bin/_expected_outputs.json. Anything that drifts
-# fails CI; anything that errors out also fails CI.
+# Runs bin/run-snippets.py --check against every snippet advertised by the
+# docs site (bin/_docs_components.py). The runner verifies each snippet
+# according to its catalog flags:
 #
-# Snippets that can't run locally (need network, a listening port, or other
-# Playground-only features) are not in the JSON and are skipped — the runner
-# only enforces matches for snippets that already have captured output.
+#   * Snippets marked runnable are executed and their stdout is compared to
+#     bin/_expected_outputs.json. Runtime failures, output drift, or new
+#     snippets without a captured baseline all fail CI.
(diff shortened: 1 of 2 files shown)


adamziel:
Closing — the strict-runner ideas here (fail on runtime errors, syntax-check non-runnable snippets, no silent skips) made sense before the #254 markdown migration, but the runner was rewritten in #254 against a different data structure (per-component markdown files with inline expected-output blocks instead of a sidecar JSON). The branch's diff no longer applies cleanly. If we want the strict-fail behavior back, easier to do it as a fresh PR against the new run-snippets.py.

PR #265, opened 2026-05-05
Close `ProcessOutputStream` in `Runtime::eval_php_code_in_subprocess`

fredrikekelund:
Codex helped me diagnose and fix this issue. 

When running the following blueprint using php-cli on Windows, I ran into an error like this:

<details>
<summary>View blueprint</summary>
<pre><code>{
	"$schema": "https://playground.wordpress.net/blueprint-schema.json",
	"landingPage": "/wp-admin/themes.php",
	"steps": [
		{
			"step": "installTheme",
			"themeData": {
				"resource": "wordpress.org/themes",
				"slug": "twentytwentyone"
			}
		},
		{
			"step": "activateTheme",
			"themeFolderName": "twentytwentyone"
		}
	]
}
</code></pre>
</details>

<details>
<summary>View logs</summary>
[...]

PR #266, opened 2026-05-07
Close import content output stream

ashfame:
Follow-up to #265.

While reviewing that PR, Codex noticed one remaining `Process::OUTPUT_FILE` consumer in `ImportContentStep::importWxr()` that also left its `ProcessOutputStream` read handle open.

This wraps the output-reading loop in `try`/`finally` and calls `close_reading()` after the importer output has been consumed.


ashfame:
@brandonpayton This one is a small change that has been sitting for a while. Can you review?

PR #267, opened 2026-05-07
Fix docs parsing of CommonMark children

ashfame:
Fixes the docs deploy failure from https://github.com/WordPress/php-toolkit/actions/runs/25510294479/job/74871014893#step:5:21.

The vendored CommonMark `Node::children()` implementation can return a plain array. `bin/build-reference.php` assumed it always returned `Traversable` and passed it directly to `iterator_to_array()`, which fails under the docs deploy PHP 8.1 job.

This adds a small `node_children()` helper that accepts either arrays or `Traversable` results, then uses it in both `bin/build-reference.php` and `bin/run-snippets.php`.

### DIFF
--- a/bin/build-reference.php
+++ b/bin/build-reference.php
@@ -115,6 +115,12 @@ function render_nodes( array $nodes ): string {
 	return rtrim( (string) $renderer->renderNodes( $nodes ) );
 }
 
+function node_children( Node $node ): array {
+	$children = $node->children();
+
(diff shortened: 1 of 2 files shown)


ashfame:
Fixed in https://github.com/WordPress/php-toolkit/commit/ba7118976a5e604ac62609ea1220fd6aeea45d31 so closing

PR #233, opened 2026-04-09
CSS: Add numeric flag to number, dimension tokens.

sirreal:
## Summary

- Implements the CSS Syntax Level 3 "type flag" (`"integer"` or `"number"`) for `<number-token>` and `<dimension-token>`, resolving the `@TODO` in `consume_numeric()`
- Adds `get_token_number_type()` accessor returning `"integer"`, `"number"`, or `null`
- `<percentage-token>` returns `null` per spec (percentages have no type flag)

## Details

The [CSS Syntax Level 3 spec](https://www.w3.org/TR/css-syntax-3/#consume-number) defines that number and dimension tokens carry a type flag indicating whether the number was written as an integer (`42`, `+7`) or a number (`42.0`, `1e2`, `.5`). This distinction matters for CSS property validation — properties like `z-index` and `column-count` require the integer type.

The flag is determined during number consumption: it defaults to `"integer"` and becomes `"number"` when a decimal point or exponent is consumed.

## Test plan

- [ ] New data-provider test with 17 cases covering integers, decimals, exponents, signed numbers, dimensions, percentages, and non-numeric tokens
- [ ] Existing tests continue to pass (no changes to `get_token_value()` behavior)

PR #297, opened 2026-06-30
CSS: Add hash token type flag

sirreal:
## Summary

Adds CSS Syntax hash-token type flag support to `CSSProcessor`.

Hash tokens now expose whether they are `id` or `unrestricted` via new constants and `get_token_type_flag()`, matching the spec more closely and giving downstream consumers the information they need to distinguish hash-token behavior.

Exposing that flag lets callers distinguish ID-selector-style hashes.

## Testing

Adds coverage for hash-token type flags, delimiter fallback, and clearing the flag between tokens.

The token type flag can likely be re-used for the number type, see
https://github.com/WordPress/php-toolkit/pull/233.

### DIFF
--- a/components/DataLiberation/CSS/class-cssprocessor.php
+++ b/components/DataLiberation/CSS/class-cssprocessor.php
@@ -131,6 +131,14 @@ class CSSProcessor {
 	public const TOKEN_FUNCTION      = 'function-token';
 
 	/**
+	 * Hash-token type flags.
+	 *
+	 * @see https://www.w3.org/TR/css-syntax-3/#typedef-hash-token
+	 */
+	public const HASH_TOKEN_ID           = 'id';
+	public const HASH_TOKEN_UNRESTRICTED = 'unrestricted';
+
+	/**
 	 * URL tokens represent unquoted URLs in url() notation.
 	 *
 	 * Valid: url(image.jpg), url(https://example.com)
@@ -214,6 +222,13 @@ class CSSProcessor {
 	 * @var string|null
 	 */
 	private $token_type = null;
+
+	/**
+	 * The type flag of the current token, if the token carries one.
+	 *
+	 * @var string|null
+	 */
+	private $token_type_flag = null;
 
 	/**
 	 * The byte offset at which the current token starts.
@@ -414,9 +429,9 @@ class CSSProcessor {
 					// Create a <hash-token>.
 					++$this->at;
 
-					// We skip this check as we don't track the type flag:
-					// > If the next 3 input code points would start an ident sequence,
-					// > set the <hash-token>'s type flag to "id".
+					$this->token_type_flag = $this->check_if_3_code_points_start_an_ident_sequence( $this->at )
+						? self::HASH_TOKEN_ID
+						: self::HASH_TOKEN_UNRESTRICTED;
 
 					// Consume an ident sequence, and set the <hash-token>'s value to the returned string.
 					$this->consume_ident_sequence();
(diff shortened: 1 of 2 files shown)


sirreal:
I've aligned this with #233, I'd like to land them more or less together.

sirreal:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

components/DataLiberation/CSS/class-cssprocessor.php
```diff
@@ -215,6 +215,13 @@ class CSSProcessor {
 	 */
 	private $token_type = null;
 
+	/**
+	 * The type flag of the current token, if the token carries one.
+	 *
+	 * @var string|null
```
dmsnell, line 221: is the set of values static and small? could we add them here instead of using `string`? or as a list in the docblock?

dmsnell:
> Hash tokens now expose whether they are id or unrestricted

can you give a brief explanation of what this is?

components/DataLiberation/CSS/class-cssprocessor.php
sirreal: Yes, this overlaps with #233 here and I avoided writing it all out here.

For now it's just the hash flags: `id` or `unrestricted`. #233 adds 2 more.

sirreal:
> can you give a brief explanation of what this is?

Goodness, I thought I documented this a bit better. #233 is more like what I had in mind with regards to documentation, the two are very linked (they'll populate the same property and expose the same method to exposed the token flags).

[This implements this part of the CSS spec:](https://www.w3.org/TR/css-syntax-3/#tokenization):

> hash tokens have a type flag set to either "id" or "unrestricted"

I believe it was necessary when using this tokenizer for selector parsing.

(#233 handles the adjacent "<number-token> and <dimension-token> additionally have a type flag set to either 'integer' or 'number'" part of the spec)

dmsnell:
so this is like `get_token_type()`? but only if the type is `hash`? and what I was getting at, is what is the meaning of `id` and `unrestricted` and why would calling code care?

even in the linked spec, I don’t see any obvious explanation for setting that flag or what it means, other than an auxiliary note that an unrestricted hash token “may not need as much escaping.”

sirreal:
> so this is like get_token_type()? but only if the type is hash?

It's like a sub-type of some tokens. Hash tokens have a flag. Number and dimension tokens as well. It conveys some additional information about the token.

Only id-flagged hash tokens are valid ID selectors, so when parsing selectors from the token stream you need to know whether you have an id or unrestricted hash token. [It's part of the grammar.](https://www.w3.org/TR/selectors-4/#grammar)

> In [`<id-selector>`](https://www.w3.org/TR/selectors-4/#typedef-id-selector), the [`<hash-token>`](https://www.w3.org/TR/css-syntax-3/#typedef-hash-token)’s value must be an [identifier](https://www.w3.org/TR/css-values-4/#css-css-identifier).

If you don't have this information from the token stream, you cannot reliably parse selectors and would need to re-parse the hash token.

dmsnell:
thanks @sirreal for the extra link. is it right then that this is more or less saying there are valid hash tokens, and there are ignored hash tokens? as in, `#\ ` is not valid in a selector because no identifier could be the space character?

sirreal:
`#000` is an example of an "unrestricted" hash token that's a relevant token and valid CSS, just not in the context of selectors:

```css
#ident {
  color: #000;
}
```

(In the case of `#\ `, that is a valid id selector! I believe because the escape sequence can start an ident.)

dmsnell:
aha, that makes so much more sense. thanks @sirreal for clarifying that. a tiny example comment would make that so much more readily-apparent when reviewing the API

sirreal:
I've added some details on the hash token type flag.

dmsnell (approved):

PR #302, opened 2026-08-07
Move processor implementations to their public paths

adamziel:
Moves the retained URL-in-text, HTML, and XML implementation files out of the `PHP/` sidecar directories and into their public processor paths.

Files changed contains exactly four 100% renames with zero additions and zero deletions. No implementation byte changes in this PR.

This PR is stacked on #304, which temporarily vacates the destination paths while retaining the public aliases. #303 contains the class declaration changes and removes those aliases.

## Testing

The Data Liberation, XML, and Markdown suites pass. PHPCS, syntax checks, direct-loader checks, and root-autoload checks pass.


adamziel:
Superseded by #303 so the review stack presents the byte-identical moves first and the declaration changes separately in #305.

PR #311, opened 2026-08-18
Prevent ZIP installs from clearing unrelated plugins and themes

### LINKED ISSUE #310: Blueprint installPlugin can delete the entire plugins directory for ZIPs rooted at ./
## Problem

`InstallPluginStep` derives `$plugin_slug` from the first ZIP entry and uses it as a recursive deletion target before invoking `Plugin_Upgrader`. A valid ZIP created with `zip -r plugin.zip .` commonly begins with `./`. The inferred slug is therefore `.`, making the target `WP_PLUGIN_DIR . "/."`. The current code recursively deletes that path, removing every existing plugin before installation.

WordPress core protects `WP_PLUGIN_DIR` inside `WP_Upgrader::install_package()`, but this pre-delete happens before core receives the package and bypasses that protection.

## Reproduction

1. Create a plugin ZIP whose entries begin with:

```text
./
./example-plugin.php
```

2. Ensure the target WordPress site already contains another plugin.
3. Run a Blueprint `installPlugin` step using that ZIP.
4. Observe that the pre-existing plugin is deleted.

## Expected behavior

Archive metadata must never resolve to the plugin root or escape it. The step should ignore `.` metadata entries when inferring a folder, validate any inferred destination as a safe single path component, and fail before filesystem mutation when no safe target can be derived. Existing unrelated plugins must remain untouched.

## Suggested regression coverage

- A ZIP beginning with `./` does not delete `wp-content/plugins`.
- `.` / `..` / absolute or separator-containing inferred targets are rejected before mutation.
- Conventional `plugin-slug/` ZIPs continue to install.
- Root-level plugin files continue through the existing filename/upgrader fallback without a pre-delete.

## AI assistance

OpenCode with `openai/gpt-5.6-sol` was used to reduce the failing archive layout, trace the destination through PHP Toolkit and WordPress core, and prepare this report. Chris Huber reviewed and remains responsible for the report.

chubes4:
A valid plugin or theme ZIP created from the current directory may begin with `./`. The old installers treated the first path segment as the target folder. That made the target `WP_PLUGIN_DIR/.` or `wp-content/themes/.` and recursively cleared unrelated plugins or themes before WordPress selected the installation folder.

This change leaves target selection, replacement, and installed-folder detection to `Plugin_Upgrader` and `Theme_Upgrader`. Regression tests cover plugin and theme archives beginning with `./` and confirm that existing unrelated plugins and themes survive.

Windows test cleanup now removes each copied WordPress site instead of retaining all copies until the job ends. Current hosted Windows runners otherwise exhaust the 15 GB temporary drive before the suite finishes.

## Testing

- `vendor/bin/phpunit components/Blueprints/Tests/Unit/Steps` (77 tests, 185 assertions)
- `vendor/bin/phpcs -d memory_limit=1G components/Blueprints/Steps/class-installpluginstep.php components/Blueprints/Steps/class-installthemestep.php components/Blueprints/Tests/Unit/Steps/InstallPluginStepTest.php components/Blueprints/Tests/Unit/Steps/InstallThemeStepTest.php components/Blueprints/Tests/Unit/Steps/StepTestCase.php`
- Each regression test failed before its source change and passed afterward.

Closes #310

### DIFF
--- a/components/Blueprints/Steps/class-installpluginstep.php
+++ b/components/Blueprints/Steps/class-installpluginstep.php
@@ -215,34 +215,6 @@ if ( ! file_exists( $plugin_zip_path ) ) {
 	exit( 1 );
 }
 
-// List files from the plugin zip
-$zip = new ZipArchive();
-if ( $zip->open( $plugin_zip_path ) !== true ) {
-	fwrite( STDERR, "Failed to open plugin zip file: " . $plugin_zip_path . "\n" );
-	exit( 1 );
-}
-
-fwrite( STDERR, "Plugin zip contents:" . "\n" );
-for ( $i = 0; $i < $zip->numFiles; $i ++ ) {
-	$filename = $zip->getNameIndex( $i );
-	$stats    = $zip->statIndex( $i );
-	$size     = $stats['size'];
-	$is_dir   = substr( $filename, - 1 ) === '/';
-}
-
-// Extract plugin slug from the zip file
-$plugin_slug = '';
-// Check the first directory in the zip file
-if ( $zip->numFiles > 0 ) {
-	$first_entry = $zip->getNameIndex( 0 );
-	// Most plugin zips have a top-level directory that is the plugin slug
-	if ( strpos( $first_entry, '/' ) !== false ) {
-		$plugin_slug = explode( '/', $first_entry )[0];
-	}
-}
-
-$zip->close();
-
 // Make sure the destination directory is writable
 $wp_plugin_dir = WP_PLUGIN_DIR;
 if ( ! is_writable( $wp_plugin_dir ) ) {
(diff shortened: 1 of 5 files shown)


chubes4:
@adamziel, could you review this Blueprint install safety fix when you have a chance? The affected step history points to you as the relevant maintainer.

The regression fixture uses a valid `./`-rooted plugin ZIP and proves installation no longer deletes an unrelated existing plugin. CI is fully green across PHP 7.2-8.5 on Linux, macOS, and Windows, plus PHPCS and docs snippets. Studio currently packages the affected `blueprints.phar`, so its downstream fix is waiting on this merge and a release.

AI assistance: OpenCode with `openai/gpt-5.6-sol` was used to trace the deletion path, implement and test the fix, and verify the CI and downstream dependency. Chris Huber reviewed and remains responsible for the change.

adamziel:
Thank you!

